PR #5, opened 2026-02-12
fix: exclusion matching, dep graph, module regex; add directory move

peteromallet:
## Summary

- **Exclusion matching false positives**: `run_grep` and `finalize_graph` matched `--exclude` patterns against absolute paths, so a pattern like `Wan2GP` would exclude ALL files when the project root was `Headless-Wan2GP`. Now matches against relative paths only.
- **Missing absolute from-imports in dep graph**: The `from_re` regex in Python's `build_dep_graph` only matched relative imports (`from .foo import bar`), missing all absolute `from X import Y` statements. This meant the dep graph was incomplete and `desloppify move` reported 0 replacements for projects using absolute imports.
- **Module regex word boundary bug**: `_has_exact_module` used `\b` which treats `.` as a word boundary, causing `source.foo` to falsely match inside `source.foo.bar`. Replaced with lookaround assertions.
- **Directory move support**: New feature — `desloppify move source/db source/core/db` moves an entire package, updating external importers while preserving intra-package relative imports.

## Test plan

- [x] `desloppify move source/constants.py source/utils/constants.py --dry-run` — shows 13 correct replacements (was 0 before fix)
- [x] `desloppify move source/db source/core/db --dry-run` — correctly moves 7 files, updates 2 external importers, preserves intra-package imports
- [x] Verified exclusion fix: `--exclude Wan2GP` no longer excludes project files from `Headless-Wan2GP/`

### DIFF
--- a/desloppify/commands/move.py
+++ b/desloppify/commands/move.py
@@ -581,24 +581,35 @@ def _cmd_move_dir(args, source_abs: str):
     moving_files = {src for src, _ in file_moves}
 
     # Compute all replacements for each file
+    # External: files outside the moved dir that import from it
     all_importer_changes: dict[str, list[tuple[str, str]]] = {}
+    # Intra-package: files inside the moved dir that use absolute imports to siblings
+    intra_pkg_changes: dict[str, list[tuple[str, str]]] = {}
+    # Self: the moved file's own relative imports pointing outside the package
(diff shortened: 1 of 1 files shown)


peteromallet:
Changes incorporated into main via cacdc6a (manually applied since codebase architecture had diverged from PR base)

PR #60, opened 2026-02-14
Fix Windows UnicodeDecodeError when reading README in status/next/scan

tobitege:
Hey Peter,

I ran the desloppify for fun against a C# / .NET repo of mine which ran into errors on my Windows encoded readme file.

On Windows with a cp1252 default text encoding, reading README.md could raise UnicodeDecodeError and crash `desloppify status`, `desloppify next`, and `desloppify scan` unless PYTHONUTF8=1 was set.

The fixes are pretty straightfoward and only involve 2 command parameters.

### DIFF
--- a/desloppify/commands/scan.py
+++ b/desloppify/commands/scan.py
@@ -313,7 +313,9 @@ def cmd_scan(args):
                 readme_path = PROJECT_ROOT / readme_name
                 if readme_path.exists():
                     try:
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks for this @tobitege! Will review in the next batch

peteromallet:
Actually very simple to just do this one by hand, thanks!

PR #131, opened 2026-02-23
Fix debug-logs fixer deleting logger wrapper methods

### LINKED ISSUE #126: debug-logs fixer incorrectly removes application logger methods
## Bug

The `debug-logs` fixer removes lines that contain `console.log` even when they are part of a structured application logger, not debug statements.

## Reproduction

Given a logger object like:

```typescript
const log = {
  info: (msg: string) => console.log(`[${ts()}] INFO  ${msg}`),
  warn: (msg: string) => console.log(`[${ts()}] WARN  ${msg}`),
  error: (msg: string) => console.error(`[${ts()}] ERROR ${msg}`),
};
```

Running `desloppify fix debug-logs` removes the `info` and `warn` lines because they use `console.log`, leaving only the `error` method. This breaks the build since `log.info(...)` and `log.warn(...)` are called throughout the codebase.

## Expected behavior

The fixer should recognize that `console.log` calls inside a logger/wrapper object are intentional production logging, not debug statements. It should not remove lines where `console.log` is wrapped in a named function/method that serves as the application's logging interface.

## Environment

- desloppify scan detected these as "tagged debug logs"
- 2 instances found, 1 was applied (1 skipped as "complex pattern")
- The applied fix broke the TypeScript build with 39 errors (`Property 'info' does not exist on type...`)

peteromallet:
## Summary
- prevent `fix debug-logs` from removing `console.*` calls that are part of named logger wrappers
- add context guards for inline arrow wrappers and method-body wrappers
- add regression tests for logger object arrow methods and method bodies

Fixes #126

## Validation
- `pytest -q desloppify/languages/typescript/tests/test_ts_fixers.py`
- `ruff check desloppify/languages/typescript/fixers/logs.py desloppify/languages/typescript/tests/test_ts_fixers.py`

### DIFF
--- a/desloppify/languages/typescript/fixers/logs.py
+++ b/desloppify/languages/typescript/fixers/logs.py
@@ -3,14 +3,105 @@
 import re
 import sys
 
+from desloppify.file_discovery import rel
 from desloppify.languages.typescript.fixers.common import (
     apply_fixer,
     collapse_blank_lines,
     extract_body_between_braces,
     find_balanced_end,
 )
-from desloppify.file_discovery import rel
 from desloppify.utils import colorize
+
+_LOGGER_WRAPPER_NAMES = frozenset(
+    {
+        "log",
+        "logger",
+        "info",
+        "warn",
+        "warning",
+        "error",
+        "debug",
+        "trace",
+        "fatal",
+        "notice",
+    }
+)
+
+_INLINE_WRAPPER_PATTERNS = (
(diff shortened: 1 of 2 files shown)


peteromallet:
Fix verified and ready to merge:\n- protects logger wrapper methods from  removal\n- regression tests added\n- CI passed all required checks\n\nMerging to main now.

PR #177, opened 2026-03-01
fix: guard detail dict access in render.py and suppress repeated plugin load warnings

### LINKED ISSUE #172: Intermittent plugin loader circular-import errors in editable install
## Summary
In an editable local install, `desloppify` intermittently emitted repeated language plugin load failures before normal command output.

This may have been caused by a transient local state during active refactor work, but the failure mode is noisy and makes it hard to trust command results.

## Repro (observed)
From a project using local editable install:

```bash
python3 -m pip install -e /Users/peteromalley/Documents/desloppify
cd /Users/peteromalley/Documents/reigh
desloppify status
# (also seen on: desloppify plan / desloppify next / desloppify show ...)
```

## Actual
Before normal output, many repeated errors appear (one per language plugin), e.g.:

```text
Language plugin .clojure failed to load: ImportError: cannot import name 'generic_lang' from partially initialized module 'desloppify.languages._framework.generic' (most likely due to a circular import)
```

Also observed for `.cxx`, `.elixir`, `.erlang`, `.java`, `.javascript`, `.rust`, etc.

## Expected
No circular-import plugin loader failures should be emitted for unrelated language plugins during standard commands.

## Notes
- This was observed in local dev while iterating on Desloppify itself.
- Later in the same session (after reinstall/update), the issue was not reproduced.
- So this might be transient/state-dependent, but worth hardening around plugin initialization and import ordering.

## Environment
- macOS
- `gh version 2.86.0`
- Desloppify installed via editable local path (`pip install -e ...`)

### LINKED ISSUE #176: Crash in desloppify next: AttributeError in render.py
## Summary
`desloppify next` can crash with:

`AttributeError: 'str' object has no attribute 'get'`

in `app/commands/next_parts/render.py` during terminal rendering.

## Version / Environment
- desloppify: `0.8.0`
- Python: `3.14.3`
- OS: macOS (arm64)

## Repro
1. Run a scan in a repo with mixed findings:
   - `desloppify scan --path .`
2. Run next with a larger count (or plain `next` in some states):
   - `desloppify next --count 15`

Observed: command prints some items, then crashes.

## Actual Error
```text
Traceback (most recent call last):
  ...
  File ".../desloppify/app/commands/next_parts/render.py", line 217, in _render_item
    if detail.get("lines"):
       ^^^^^^^^^^
AttributeError: 'str' object has no attribute 'get'
```

I can reproduce this repeatedly when `next` iterates over certain findings/groups (often after large review/plan state changes).

## Expected
`desloppify next` should render all queued items (or gracefully skip malformed details), not crash.

## Suspected Root Cause
`_render_item` appears to assume `detail` is always a dict. In this case it can be a string, so `detail.get(...)` raises.

## Suggestion
- Guard the render path with type checks before dict access (e.g. `isinstance(detail, dict)`).
- Optionally normalize finding detail shapes earlier so terminal rendering receives a consistent schema.
- Add a regression test where an item has string detail payload.

xliry:
## Summary

Fixes two reported issues:

**Fix #176 — AttributeError crash in `desloppify next`**

`_render_item` in `app/commands/next_parts/render.py` assumes `item['detail']` is always a dict, but it can be a string when findings come from certain scan states. This causes:
```
AttributeError: 'str' object has no attribute 'get'
```

Added a `_item_detail(item)` helper that always returns a dict (returning `{}` for strings/None/missing), and replaced all four `item.get("detail", {})` call sites with it.

**Fix #172 — Repeated plugin load-failure warnings on every command**

Plugin load errors were re-logged on every subsequent `load_all()` call. This caused noisy output like:
```
Language plugin .clojure failed to load: ImportError: cannot import name 'generic_lang' from partially initialized module ...
```
appearing before normal command output on every invocation.

Changes:
- Removed the `_report_load_errors_for_load_all()` call from the early-return path in `load_all()` — errors are now logged only once (on the initial load)
- Added a second-pass retry for plugins that failed due to circular-import errors — after the first pass all framework modules (e.g. `generic.py`) are fully initialised, so the retry typically succeeds

## Test plan
- [ ] `desloppify next` no longer crashes with `AttributeError` when a finding has a string `detail` field
- [ ] Plugin load warnings appear at most once per process
- [ ] `python3 -m pytest desloppify/tests/lang/common/test_lang_discovery.py` — all 4 tests pass

Fixes #176
Fixes #172

### DIFF
--- a/desloppify/app/commands/next_parts/render.py
+++ b/desloppify/app/commands/next_parts/render.py
@@ -80,6 +80,12 @@ def is_auto_fix_command(command: str | None) -> bool:
     return cmd.startswith("desloppify fix ") and "--dry-run" in cmd
 
 
+def _item_detail(item: dict) -> dict:
+    """Return item['detail'] as a dict, normalising strings to an empty dict."""
+    d = item.get("detail")
+    return d if isinstance(d, dict) else {}
+
+
 def _effort_tag(item: dict) -> str:
     """Return a short effort/type tag for a queue item."""
     if item.get("detector") == "review":
@@ -173,7 +179,7 @@ def _render_item(
 
     kind = item.get("kind", "finding")
     if kind == "subjective_dimension":
-        detail = item.get("detail", {})
+        detail = _item_detail(item)
         subjective_score = float(
             detail.get("strict_score", item.get("subjective_score", 100.0))
         )
@@ -213,7 +219,7 @@ def _render_item(
     print(f"  File: {item.get('file', '')}")
     print(colorize(f"  ID:   {item.get('id', '')}", "dim"))
 
-    detail = item.get("detail", {})
+    detail = _item_detail(item)
     if detail.get("lines"):
         print(f"  Lines: {', '.join(str(line_no) for line_no in detail['lines'][:8])}")
     if detail.get("category"):
@@ -271,7 +277,7 @@ def _render_item(
         # Review findings: show dimension drag from breakdown
         try:
             from desloppify.scoring import compute_health_breakdown
-            dim_key = item.get("detail", {}).get("dimension", "")
+            dim_key = _item_detail(item).get("dimension", "")
             if dim_key:
                 breakdown = compute_health_breakdown(dim_scores)
                 for entry in breakdown.get("entries", []):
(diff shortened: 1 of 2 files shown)


peteromallet:
Closing stale PR — the render.py guard and plugin warning fixes have been addressed in subsequent releases. Thank you @xliry.

Issue #24, opened 2026-02-12
detector: hook return bloat (hooks returning too many values)


peteromallet:
## Gap

Desloppify catches bloated \`*Props\` interfaces (>14 fields) and monster functions (>150 LOC), but misses a critical React-specific pattern: **hooks that return too many values**.

A hook returning 50+ fields is functionally identical to a component receiving 50+ props — it creates tight coupling, makes consumers hard to understand, and signals that the hook is doing too many things. But because hooks return objects (not interfaces named \`*Props\`), the props detector doesn't see them.

Real-world examples from a 1000-file React codebase:

| Hook | Return fields | What it does |
|------|--------------|--------------|
| \`useImageGenForm\` | 50+ | Prompts, references, LoRAs, UI state, submission, settings — everything |
| \`useSharedLightboxState\` | 40+ (8 nested objects) | Variants, navigation, shots, lineage, stars, refs, buttons |
| \`usePaneState\` | 18 | Pane visibility, animations, positions, callbacks |
| \`useVideoEditing\` | 30+ | Portions, LoRAs, settings, tasks, validation |

These hooks become "God hooks" — the hook equivalent of God classes. They're the primary source of complexity in the codebase.

## Why it's general

Custom hooks are the primary abstraction mechanism in modern React. As features grow, hooks accumulate return values. Unlike class methods (which the \`gods\` detector catches), hook return bloat is invisible to current detectors.

The pattern is universal:
1. Hook starts small: \`return { value, setValue }\`
2. Features accrete: \`return { value, setValue, isLoading, error, reset, validate, ... }\`
3. Eventually: \`return { /* 50 fields */ }\`
4. Nobody refactors because all consumers depend on the monolithic return

## Detection

New detector: \`detect_hook_return_bloat()\`

**Algorithm:**
1. Find custom hook functions (functions whose name starts with \`use\` and are exported, or files named \`use*.ts\`)
2. Find the main return statement
3. Count top-level keys in the returned object literal
4. For hooks returning nested objects, count total leaf fields
5. Flag if count exceeds threshold (suggested: >20 fields)

```python
def detect_hook_return_bloat(path: Path) -> tuple[list[dict], int]:
    """Find custom hooks returning objects with too many fields."""
    files = find_ts_files(path)
    entries = []
    total_hooks = 0
    
    # Find hook functions (exported functions starting with 'use')
    hook_pattern = re.compile(
[...]

Issue #22, opened 2026-02-12
New detector: param_naming_consistency — catch inconsistent parameter names across similar functions


peteromallet:
## Problem

When multiple functions across a codebase serve similar purposes (e.g., processing video files, handling task IDs), they often name their parameters differently. This creates cognitive load, makes the codebase feel ad-hoc, and causes subtle bugs when developers assume parameter names are consistent.

This is a **cross-file semantic consistency** problem that no existing linter detects — it requires understanding that two parameters in different files serve the same purpose.

## Real-World Example

In [Headless-Wan2GP](https://github.com/peteromallet/Headless-Wan2GP), extensive parameter naming inconsistencies were found:

### Video input paths (same semantic purpose, 3 different names):

| Function | File | Parameter Name |
|----------|------|---------------|
| `apply_brightness_to_video_frames()` | `video_transforms.py` | `input_video_path` |
| `extract_frames_from_video()` | `frame_extraction.py` | `video_path` |
| `apply_color_matching_to_video()` | `color_matching.py` | `video_path` |
| `debug_video_analysis()` | `debug_utils.py` | `video_path` |
| `reverse_video()` | `video_transforms.py` | `input_video_path` |

### Task ID (same semantic purpose, 4 different names):

| Function | File | Parameter Name |
|----------|------|---------------|
| `handle_extract_frame_task()` | `extract_frame.py` | `task_id` |
| `apply_brightness_to_video_frames()` | `video_transforms.py` | `task_id_for_logging` |
| `extract_last_frame_as_image()` | `frame_extraction.py` | `task_id_for_log` |
| `image_to_frame()` | `frame_utils.py` | `task_id_for_logging` |
| `load_pil_images()` | `frame_utils.py` | `task_id_for_log` |
| `_handle_travel_stitch_task()` | `stitch.py` | `stitch_task_id_str` |
| `_handle_travel_orchestrator_task()` | `orchestrator.py` | `orchestrator_task_id_str` |

### Image input paths (same semantic purpose, 3 different names):

| Function | File | Parameter Name |
|----------|------|---------------|
| `extract_pose_keypoints()` | `pose_utils.py` | `image_path` |
| `image_to_frame()` | `frame_utils.py` | `image_path_str` |
| `apply_strength_to_image()` | `frame_utils.py` | `image_path_input` |

### Output paths (same semantic purpose, 3 different names):

| Function | File | Parameter Name |
|----------|------|---------------|
[...]

Issue #21, opened 2026-02-12
detector: window global escape hatches


peteromallet:
## Gap

No detector catches \`window.__*\` assignments — a common anti-pattern where internal state is leaked to the global scope as a debugging/coordination escape hatch.

In a real-world 1000-file React codebase, 5 separate global assignments were found:

```typescript
window.__AUTH_MANAGER__ = authManager;           // AuthStateManager.ts
window.__PROJECT_CONTEXT__ = contextValue;       // ProjectContext.tsx
window.__REACT_QUERY_CLIENT__ = queryClient;     // TasksPane.tsx
window.__supabase_client__ = supabase;           // client.ts
window.__projectDebugLog = [];                   // mobileProjectDebug.ts
```

These indicate the architecture isn't properly routing state through React's component tree. They break testing, make refactoring harder, create invisible coupling between modules, and can leak sensitive state (auth tokens, database clients) to browser console access.

## Why it's general

Window globals are a universal escape hatch in browser codebases. They appear when:
- Debugging tools need access to internal state (\`window.__DEBUG__\`)
- Two systems need to communicate outside React's tree (\`window.__AUTH__\`)
- Legacy migration leaves global bridges in place
- Quick fixes bypass proper architecture

They're particularly insidious because they work perfectly and never break — so they're never removed. Every long-lived SPA accumulates them.

## Detection

New detector: \`detect_window_globals()\`

```python
def detect_window_globals(path: Path) -> tuple[list[dict], int]:
    """Find window.__* assignments — global state escape hatches."""
    files = find_ts_files(path)
    entries = []
    # Matches: window.__foo = ..., (window as any).__foo = ..., window['__foo'] = ...
    pattern = re.compile(
        r"""(?:window\s*(?:as\s+any\s*)\)?\s*\.\s*(__\w+)"""   # window.__name or (window as any).__name
        r"""|window\s*\[\s*['"](__\w+)['"]\s*\])"""              # window['__name']
        r"""\s*=""",                                              # followed by assignment
        re.VERBOSE,
    )
    for filepath in files:
        try:
            content = Path(filepath).read_text()
            for i, line in enumerate(content.splitlines()):
                m = pattern.search(line)
                if m:
                    name = m.group(1) or m.group(2)
                    entries.append({
                        "file": filepath, "line": i + 1,
                        "name": name, "content": line.strip()[:100],
                    })
        except (OSError, UnicodeDecodeError):
            continue
    return entries, len(files)
```

## Example output

```
[...]

Issue #20, opened 2026-02-12
New detector: layer_violation — enforce module hierarchy and prevent backward imports


peteromallet:
## Problem

Well-structured codebases have a module hierarchy where lower-level modules don't import from higher-level ones. When this hierarchy is violated, it creates:

1. **Circular dependency risk** — backward imports are the #1 cause of circular imports in Python
2. **Coupling inversion** — infrastructure becomes dependent on business logic
3. **Testing difficulty** — you can't test core modules without pulling in the entire application
4. **Unclear ownership** — the dependency direction becomes ambiguous

Desloppify's `cycles` detector catches *circular* imports, but a **layer violation can exist without creating a cycle** — it's a unidirectional dependency in the wrong direction.

## Real-World Example

In [Headless-Wan2GP](https://github.com/peteromallet/Headless-Wan2GP), the intended module hierarchy is:

```
source/core/        → Infrastructure (db, log, params, constants)
source/utils/       → Shared utilities
source/media/       → Media processing
source/models/      → Model integration
source/task_handlers/ → Task orchestration (highest level)
```

**Violation found:**

```python
# source/core/params/phase.py (line 58)
# LOWER-LEVEL module importing from HIGHER-LEVEL module
from source.task_handlers.tasks.task_conversion import parse_phase_config
```

A core infrastructure module (`source/core/params/`) imports from a task handler (`source/task_handlers/tasks/`). This means you can't use the params module without the entire task handling system — defeating the purpose of having a separate core layer.

In this case the import was deferred (inside a method, wrapped in try/except) specifically to avoid circular import issues — which is itself a code smell indicating the layering is wrong. The function `parse_phase_config` should either live in `core/params/` or the dependency should be inverted.

### Configuration

Users would define their module hierarchy in a config file (e.g., `.desloppify/layers.yaml`):

```yaml
layers:
  - name: core
    paths: ["source/core/"]
    level: 0  # lowest level
  - name: utils
    paths: ["source/utils/"]
    level: 1
  - name: media
    paths: ["source/media/"]
    level: 2
  - name: models
    paths: ["source/models/"]
    level: 3
  - name: handlers
    paths: ["source/task_handlers/"]
    level: 4  # highest level
```

### Algorithm

1. For each Python file, determine its layer based on file path
2. Parse all import statements (AST — already done in Phase 1/3)
3. Resolve import targets to file paths (already done for cycles/orphaned)
4. If an import goes from a lower-level layer to a higher-level layer, flag it
[...]

PR #178, opened 2026-03-02
PHP deep integration: test coverage, security, imports, tree-sitter

ikx94:
## Summary

**WIP** — coded with [Claude Code](https://claude.com/claude-code).

Upgrades PHP from a minimal phpstan-only plugin to near-parity with Python/TypeScript for framework-aware analysis. Addresses ~20 distinct gaps across security detection, import resolution, test coverage mapping, and tree-sitter queries.

The prompt that started this:

> Desloppify has a hook system for language-specific test coverage, but PHP registers zero hooks. Let me look at reference implementations and the PHP plugin.
>
> Three bugs/gaps:
>
> 1. No PHP test_coverage hooks — languages/php/__init__.py doesn't pass test_coverage_module, so no PHP-specific test-to-source mapping, no assertion patterns, nothing.
> 2. _infer_lang_name doesn't know .php — even if hooks existed, the fallback lang inference in mapping.py:26-34 has no .php entry, so PHP files would never trigger the hooks.
> 3. No entry_patterns — generic_lang hardcodes entry_patterns=[] at line 250, so all Laravel controllers/commands/middleware/providers are flagged as orphaned (0 importers).
>
> [... full 22-issue audit followed, covering critical/high/medium/low gaps ...]

### New files
- **`languages/php/test_coverage.py`** — PHPUnit/Pest assertion patterns, `use`-statement parsing (incl group `use A\{B,C}`), `has_testable_logic` (excludes interfaces/config/migration stubs), `is_runtime_entrypoint` (controllers/commands/jobs/routes), `map_test_to_source`, `strip_test_markers`, `strip_comments`
- **`languages/php/review_data/dimensions.override.json`** — PHP/Laravel abstraction fitness review guidance

### Fixes (12 files, ~20 issues)

| Issue | File | Fix |
|-------|------|-----|
| #14 SECRET_NAME_RE misses `$var =` | `security.py` | Add `\$?` to char class |
| #11 ENV_LOOKUPS missing PHP | `security.py` | Add `env(`, `getenv(`, `$_ENV[`, `config(` |
| #13 LOG_CALLS missing PHP | `security.py` | Add `Log::info/error/...`, `error_log()` |
| #12 RANDOM_CALLS missing PHP | `security.py` | Add `rand(`, `mt_rand(` |
| #5 Import query misses group use | `_specs.py`, `_imports.py` | New query pattern + prefix prepending |
| #4 resolve_php_import ignores composer.json | `_imports.py` | Read PSR-4 mappings from `composer.json` |
| #6 Aliased `use Foo as Bar` FP | `_unused_imports.py` | Extract alias name, search for alias |
| #8 class_query misses enums | `_specs.py` | Add `enum_declaration` |
| #9 Closure node type wrong | `_complexity.py` | Add `anonymous_function_creation_expression` |
| #10 Match arms not counted | `_complexity.py` | Add `match_conditional_expression` |
| #22 Signature allowlist | `signature.py` | Add `__construct`, `boot`, `register`, `render`, `toArray`, `rules`, `authorize` |
| #2 _infer_lang_name | `mapping.py` | Add `.php` → `php` |
| #3 entry_patterns hardcoded empty | `generic.py` | New `entry_patterns` param |
| #1,#16,#19,#20,#24 PHP plugin | `php/__init__.py` | Wire test_coverage, entry_patterns, zone_rules, excludes |
| #15 Remediation text | `rules.py` | Add `random_bytes()` for PHP |
| #21 Review override | `dimensions.override.json` | PHP/Laravel-specific guidance |

### Deferred
- **#17** barrel_names — N/A for PHP
- **#18** dynamic_import_finder for service container — entry_patterns cover most cases

## Test plan
- [x] All 3305 tests pass (3305 passed, 5 skipped)
- [x] Smoke-tested: import parsing, entrypoint detection, secret matching, test mapping
- [ ] Test against a real Laravel codebase
- [ ] Verify tree-sitter queries against edge cases (nested group use, multi-alias)

### DIFF
--- a/desloppify/languages/php/__init__.py
+++ b/desloppify/languages/php/__init__.py
@@ -26,30 +26,42 @@ PHP_ZONE_RULES = [
 # ── Entry patterns (files legitimately having zero importers) ─
 
 PHP_ENTRY_PATTERNS = [
-    # Laravel runtime entrypoints
-    "/routes/",
-    "/app/Http/Controllers/",
-    "/app/Console/Commands/",
-    "/app/Http/Middleware/",
-    "/app/Providers/",
-    "/app/Jobs/",
-    "/app/Listeners/",
-    "/app/Mail/",
-    "/app/Notifications/",
-    "/app/Policies/",
-    "/app/Events/",
-    "/app/Observers/",
-    "/app/Rules/",
-    "/app/Casts/",
+    # Laravel runtime entrypoints (no leading / — rel() returns relative paths)
+    "routes/",
+    "app/Http/Controllers/",
+    "app/Http/Middleware/",
+    "app/Http/Requests/",
+    "app/Console/Commands/",
+    "app/Providers/",
+    "app/Jobs/",
+    "app/Listeners/",
+    "app/Mail/",
+    "app/Notifications/",
+    "app/Policies/",
+    "app/Events/",
+    "app/Observers/",
+    "app/Rules/",
+    "app/Casts/",
+    "app/Exceptions/",
+    # Convention-loaded by Laravel / packages (zero explicit importers)
+    "app/Models/",
+    "app/Enums/",
+    "app/Actions/",
+    "app/Filament/",
+    "app/Livewire/",
+    "app/View/Components/",
     # Test files
-    "/tests/",
+    "tests/",
     "Test.php",
-    # Config / bootstrap
-    "/config/",
-    "/database/migrations/",
-    "/database/seeders/",
-    "/database/factories/",
-    "/resources/views/",
+    # Config / bootstrap / public
+    "config/",
+    "database/migrations/",
+    "database/seeders/",
+    "database/factories/",
+    "resources/views/",
+    "bootstrap/",
+    "public/",
+    "lang/",
 ]
 
 # ── Plugin registration ──────────────────────────────────────
--- a/desloppify/languages/php/test_coverage.py
+++ b/desloppify/languages/php/test_coverage.py
@@ -161,18 +161,23 @@ def has_testable_logic(filepath: str, content: str) -> bool:
 _ENTRYPOINT_PATH_PATTERNS = [
     "routes/",
     "app/http/controllers/",
-    "app/console/commands/",
     "app/http/middleware/",
+    "app/http/requests/",
+    "app/console/commands/",
     "app/providers/",
     "app/jobs/",
     "app/listeners/",
     "app/mail/",
     "app/notifications/",
     "app/policies/",
+    "app/events/",
+    "app/observers/",
+    "app/filament/",
+    "app/livewire/",
 ]
 
 _ENTRYPOINT_CONTENT_PATTERNS = [
-    re.compile(r"class\s+\w+\s+extends\s+(?:Controller|Command|Job|Mailable|Notification)\b"),
+    re.compile(r"class\s+\w+\s+extends\s+(?:Controller|Command|Job|Mailable|Notification|FormRequest|Resource)\b"),
     re.compile(r"class\s+\w+\s+implements\s+ShouldQueue\b"),
     re.compile(r"Route::\w+\("),
     re.compile(r"Artisan::command\("),


peteromallet:
Hey, nice work on this! Looks like a solid implementation.

I don't have any PHP projects to test against on my end — would you be able to run it against a real Laravel codebase and verify those tree-sitter edge cases you mentioned in the test plan? That would go a long way toward getting this merged.

ikx94:
Tested against two real Laravel codebases — a large e-commerce app (~500 production files) and a SaaS booking platform (~760 files). Found and fixed a bug that was tanking PHP scores everywhere.

### The bug

All `PHP_ENTRY_PATTERNS` had a leading `/` (e.g. `/app/Http/Controllers/`), but the orphaned detector does a raw substring check against `rel()` output which returns paths *without* a leading slash (`app/Http/Controllers/Foo.php`). The check `"/app/Http/Controllers/" in "app/Http/Controllers/Foo.php"` always fails — **zero entry patterns were matching**. Every Laravel file was flagged orphaned regardless of config.

This is why code quality was stuck at ~50% on every Laravel project.

### The fix (just pushed)

Stripped leading `/` from all 24 patterns. Also added missing Laravel convention-loaded directories that legitimately have zero explicit importers:

- `app/Http/Requests/` (injected by Laravel DI)
- `app/Exceptions/`
- `app/Models/` (Eloquent, route model binding, relationship references)
- `app/Enums/` (model casts, Filament forms)
- `app/Actions/` (Fortify/Jetstream/custom, bound in service providers)
- `app/Filament/` (auto-discovered by Filament)
- `app/Livewire/` (auto-discovered by Livewire)
- `app/View/Components/` (Blade component auto-discovery)
- `bootstrap/`, `public/`, `lang/`

Extended `test_coverage.py` entrypoint detection similarly — added `events/`, `observers/`, `filament/`, `livewire/`, `http/requests/` path patterns and `FormRequest`/`Resource` to content-based extends matching.

### Real-world results

**Project A** (large e-commerce, ~500 production files):
- Code quality: 51.2% → 97.7%
- Orphaned files: 465 → ~20
- Overall: 87.1 → 90.1

**Project B** (SaaS platform, ~760 files):
- Code quality: 55.5% → 97.5%
- Orphaned files: 103 → 4 (3 are `_ide_helper` generated files, 1 genuinely unused service)
- Overall: 80.1 → 82.8

### Note: same entry pattern bug exists in other languages

Dart, Go, TypeScript, C#, GDScript all have leading `/` in their entry patterns too. Zone rules aren't affected (they use `_match_pattern` which pads the path). Impact is lower on those languages since they have fewer convention-based entry points, but it's still dead code. Worth a follow-up PR.

### Tree-sitter edge cases

The tree-sitter queries handled everything I threw at them — standard `use` statements, group `use A\B\{C, D}`, aliased imports, trait use declarations. PSR-4 resolution via composer.json worked correctly across both projects. No edge case failures found.

peteromallet:
Really appreciate the thoroughness here — the real-world testing against two Laravel codebases is exactly what was needed, and the entry pattern slash bug is a great catch. The before/after numbers speak for themselves (51% → 98% code quality is huge).

I'm in the middle of pushing a big batch of changes right now so I don't want to merge this on top of that and create a mess. Going to add this in once things settle down on my end. Thanks again for the work on this — it's a solid contribution.

iNem0o:
@ikx94 Great work here, and very convincing Laravel validation! 

Drive-by comment from a newcomer: I only discovered this project recently, so my understanding is still limited.

This PR introduces many Laravel-specific conventions into the PHP layer. My concern is that others framework projects could get noisier results (false positives/negatives on entrypoints and architecture signals) if these conventions are applied by default.

Would moving to a framework-agnostic PHP design in this PR require a larger refactor, or would you prefer keeping this Laravel-first now and adding framework support for PHP in a follow-up?

I’m not trying to block this PR, I’m trying to align on direction early so I can try to contribute Symfony support later :)

ikx94:
> @ikx94 Great work here, and very convincing Laravel validation!
> 
> Drive-by comment from a newcomer: I only discovered this project recently, so my understanding is still limited.
> 
> This PR introduces many Laravel-specific conventions into the PHP layer. My concern is that others framework projects could get noisier results (false positives/negatives on entrypoints and architecture signals) if these conventions are applied by default.
> 
> Would moving to a framework-agnostic PHP design in this PR require a larger refactor, or would you prefer keeping this Laravel-first now and adding framework support for PHP in a follow-up?
> 
> I’m not trying to block this PR, I’m trying to align on direction early so I can try to contribute Symfony support later :)

I would encorage you to try it on Symfony and to make it less Laravel specific, that’s definitely the way to go

peteromallet:
Merged manually with rebase to resolve conflicts — thanks @ikx94 for the comprehensive PHP upgrade! This brings PHP to near-parity with Python/TypeScript. Author credit preserved in commit.

PR #205, opened 2026-03-04
fix(config): deterministic and transactional state->config migration

juzigu40-ui:
## Summary
This PR hardens legacy `state -> config` bootstrap migration in `desloppify/base/config.py` to remove two reliability hazards present in snapshot `6eb2065`:
- non-deterministic scalar precedence across multiple state files
- destructive partial migration when `config.json` persistence fails

## What changed
- Added deterministic migration order via `_state_files_for_migration()`.
  - `state-*.json` processed in stable lexical order
  - `state.json` processed after scoped files
- Switched migration to a two-phase flow:
  - Phase 1: merge legacy config in-memory only
  - Phase 2: write `config.json`; only after successful write, strip `config` from source state files
- Updated `_migrate_single_state_file()` to be non-destructive (merge-only)

## Why this matters
Before this patch, migration could rewrite source state files before destination durability was guaranteed. On write failure (permissions, transient I/O, full disk), original embedded config could be lost while `config.json` was still missing. This patch prevents that failure mode.

It also stabilizes scalar precedence so equivalent inputs do not produce environment-dependent outcomes from filesystem iteration order.

## Tests
Added regression tests in `desloppify/tests/core/test_config.py`:
- `test_scalar_merge_order_is_deterministic`
- `test_save_failure_does_not_strip_source_state_config`

## Local verification
- `uv run python -m pytest -q desloppify/tests/core/test_config.py`
- `uv run python -m ruff check desloppify/base/config.py desloppify/tests/core/test_config.py`

### DIFF
--- a/desloppify/base/config.py
+++ b/desloppify/base/config.py
@@ -377,8 +377,14 @@ def _migrate_single_state_file(config: dict, path: Path) -> bool:
         return False
 
     _merge_legacy_state_config(config, old_config)
-    _strip_config_from_state_file(path, state_data)
     return True
+
+
+def _state_files_for_migration(state_dir: Path) -> list[Path]:
+    """Return state files in deterministic migration order."""
+    scoped = sorted(state_dir.glob("state-*.json"), key=lambda p: p.name)
+    root = sorted(state_dir.glob("state.json"), key=lambda p: p.name)
+    return [*scoped, *root]
 
 
 def _migrate_from_state_files(config_path: Path) -> dict:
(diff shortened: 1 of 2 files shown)


juzigu40-ui:
@peteromallet Added two regression tests that capture the practical impact discussed in #204:

1. deterministic scalar merge precedence across multiple state files
2. no source-state config stripping if config persistence fails

This keeps migration recoverable under write failure and deterministic across environments.

PR #232, opened 2026-03-05
Found a way to game the scoring system

ShawTim:
While reviewing the codebase for issue #204, I found that the scoring system's floor blending mechanism (lines 17-22 in `scoring.py`) contradicts the README's claim that "the scoring resists gaming."

The 30% floor weight means historical cleanliness masks current problems. A codebase with critical bugs can still score 60+ just because it was clean last week.

Details in the attached report. This directly addresses the "poorly engineered" challenge in issue #204.

### DIFF
--- a/SCORING_FLAW_REPORT.md
+++ b/SCORING_FLAW_REPORT.md
@@ -1,0 +1,63 @@
+# Scoring Floor Logic Allows "Outlier Masking" via File Merging
+
+While the `floor` mechanism in `desloppify/app/commands/review/batch/scoring.py` is intended to be an anti-gaming feature (preventing high averages from masking outliers), its implementation using `min(score_raw_by_dim)` introduces a direct gaming vector.
+
+## The Intended Mechanism
+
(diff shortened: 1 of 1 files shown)


ShawTim:
Good catch on the previous analysis. After re-auditing, the core issue isn't about historical data—it's about how the `min(score_raw_by_dim)` floor can be bypassed through file merging.

The mathematical proof is in the updated report: merging a terrible file into a large clean file raises the score from 63.6 to 90.9 without fixing any actual code.

This is a legitimate gaming vector against the "scoring resists gaming" claim. Ready for review.

peteromallet:
Thanks @ShawTim for the thorough analysis of the scoring system! The floor-penalty gaming scenario you identified is a real insight. This is better suited as a GitHub issue for tracking rather than a PR with a report file — would you mind filing it as an issue so we can reference it when hardening the scoring logic? Closing the PR but the analysis is appreciated.

ShawTim:
done, created #379

PR #234, opened 2026-03-05
Fix Python 3.10 Compatibility

sungdark:
This PR fixes Python 3.10 compatibility issues by removing the NotRequired type annotation introduced in Python 3.11 and provides backward compatibility support for older Python versions.


peteromallet:
Thanks @sungdark for looking into Python 3.10 compatibility\! Unfortunately this approach would replace the import-boundary governance code in `compatibility.py` that's still actively used on main. If we revisit 3.10 support, we'd want to add the backports alongside the existing governance code rather than replacing it. Appreciate the effort though\!

PR #239, opened 2026-03-06
fix: emergency plan reconciliation on save failure after resolve

xliry:
## Summary

- On plan-save failure in `_update_living_plan_after_resolve`, reloads plan from disk, applies `reconcile_plan_after_scan` to supersede now-resolved IDs, then retries save — instead of just printing a warning
- Adds `state` parameter to `_update_living_plan_after_resolve`; call site updated accordingly
- Adds bounty verdict for S315 (@DavidBuchanan314): real gap confirmed (non-atomic state+plan writes), severity overstated, reconciliation-direction claim incorrect

## Problem

After a successful `resolve`, state is saved first (line 180), then plan is updated separately. If `save_plan` raised an OSError, the code caught it and warned the user — but left the plan pointing to issues that were already resolved in state. While the next `desloppify scan` would self-heal via `reconcile_plan_after_scan`, there was an unnecessary desync window with no immediate recovery.

## Fix

Wrapped the `save_plan(plan)` call in its own inner try/except. On failure:
1. Reload plan fresh from disk
2. Call `reconcile_plan_after_scan(fresh_plan, state)` — supersedes the now-resolved IDs
3. Retry `save_plan(fresh_plan)`
4. Only show warning if recovery also fails

## Test plan
- [ ] Verify `cmd_resolve` passes `state` correctly to `_update_living_plan_after_resolve`
- [ ] Verify import of `reconcile_plan_after_scan` from `desloppify.engine.plan` is valid
- [ ] Verify normal resolve path is unchanged (no recovery triggered when save succeeds)

### DIFF
--- a/bounty-verdicts/DavidBuchanan314-4002034163.json
+++ b/bounty-verdicts/DavidBuchanan314-4002034163.json
@@ -1,0 +1,39 @@
+{
+  "submitter": "DavidBuchanan314",
+  "comment_id": 4002034163,
+  "verdict": "YES_WITH_CAVEATS",
+  "verdict_date": "2026-03-06",
+  "scoreboard_id": "S315",
+  "scores": {
+    "signal": 4,
+    "originality": 4,
+    "core_impact": 2,
+    "overall": 3
+  },
+  "claims": [
+    {
+      "id": "claim-1",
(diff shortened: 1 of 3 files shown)


xliry:
Closing — test PR from pipeline development.

PR #241, opened 2026-03-06
fix: skip suppressed issues in auto_resolve_disappeared (#400)

xliry:
## Summary

- **Bounty verification:** @yv-was-taken (comment 4005041585) — YES WITH CAVEATS
- **Flaw confirmed:** `_mark_auto_resolved()` unconditionally clears `suppressed`/`suppressed_at`/`suppression_pattern` for all issues, including suppressed ones. `auto_resolve_disappeared()` had no guard to skip suppressed issues before calling it.
- **Fix:** Add `if previous.get("suppressed"): continue` after the suspect-detector guard in `auto_resolve_disappeared()`.

## What was wrong

`_mark_auto_resolved()` (`merge_issues.py:51-55`) sets:
```python
issue["suppressed"] = False
issue["suppressed_at"] = None
issue["suppression_pattern"] = None
```
unconditionally. A suppressed open issue outside the current `scan_path` would have its suppression cleared by the out-of-scope auto-resolve path (lines 104–116). The same applied to suppressed issues that genuinely disappeared from scan.

## Fix

Add one guard in `auto_resolve_disappeared()` immediately after the suspect-detector check:

```python
if previous.get("suppressed"):
    continue
```

Suppressed issues were explicitly opted out of reporting via ignore patterns. A scan should not auto-resolve them — the ignore pattern is the authoritative source for suppression decisions.

## Notes on the claim

The "permanent suppression loss" characterization is overstated: on the next in-scope scan where the issue reappears and the ignore pattern still matches, `upsert_issues()` re-applies suppression. However, a secondary inconsistency exists (not identified in the submission): re-suppression via `upsert_issues()` hits `continue` before the status-reopen check, leaving the issue in `{status: auto_resolved, suppressed: True}` — a mixed state.

## Test plan

- All `desloppify/tests/state/` tests pass (88 passed)
- Suppression/merge tests pass (133 passed)

### DIFF
--- a/bounty-verdicts/@yv-was-taken-4005041585.json
+++ b/bounty-verdicts/@yv-was-taken-4005041585.json
@@ -1,0 +1,38 @@
+{
+  "submitter": "yv-was-taken",
+  "comment_id": 4005041585,
+  "verdict": "YES_WITH_CAVEATS",
+  "verdict_date": "2026-03-06",
+  "scores": {
+    "signal": 5,
+    "originality": 5,
+    "core_impact": 1,
+    "overall": 3
+  },
+  "claims": [
+    {
+      "id": "claim-1",
+      "title": "_mark_auto_resolved() clears suppression fields unconditionally for all issues including suppressed ones",
+      "status": "CONFIRMED",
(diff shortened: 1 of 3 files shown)


xliry:
Closing — test PR from pipeline development.

PR #252, opened 2026-03-06
feat: bounty verdict @agustif — upgrade_plan_to_v7 version downgrade

xliry:
**Submission:** https://github.com/peteromallet/desloppify/issues/204#issuecomment-4000584288
**Author:** @agustif

## Problem (in our own words)

load_plan() (persistence.py:67) calls ensure_plan_defaults() unconditionally for all plans. ensure_plan_defaults() (schema.py:198) always calls upgrade_plan_to_v7(), which at schema_migrations.py:304-305 sets plan['version'] = V7_SCHEMA_VERSION without checking whether the original version was already higher. A plan saved with version 8+ loaded by load_plan() has its in-memory version silently forced to 7; any subsequent save_plan() call (which also calls ensure_plan_defaults() at persistence.py:80) permanently writes version 7 to disk, downgrading the on-disk plan.

## Evidence

- `schema_migrations.py:304-305` — `if plan.get("version") != V7_SCHEMA_VERSION: plan["version"] = V7_SCHEMA_VERSION` — unconditional version clobber in upgrade_plan_to_v7()
- `persistence.py:67` — `ensure_plan_defaults(data)` called for all plans including version > PLAN_VERSION
- `persistence.py:80` — `ensure_plan_defaults(plan)` in save_plan() means every save re-runs the downgrade
- `schema.py:198` — `_upgrade_plan_to_v7(plan)` called unconditionally inside ensure_plan_defaults()
- `persistence.py:59-65` — warning is printed for version > PLAN_VERSION but does not mention the impending version downgrade on save

## Fix

In upgrade_plan_to_v7() (schema_migrations.py), the unconditional plan['version'] = V7_SCHEMA_VERSION at the end of the function is changed to only apply when original_version < V7_SCHEMA_VERSION. Plans with version > V7_SCHEMA_VERSION now retain their version number instead of being silently downgraded. The warn-on-load behavior in load_plan() is preserved.

## Verdict

| **Is this poor engineering?** | YES | A function named upgrade_plan_to_v7() should only act on plans with version < 7. The missing version guard turns what should be a one-directional upgrade into an unconditional version clobber that silently downgrades newer plans on save. |
| **Is this at least somewhat significant?** | NO | core_impact is 2/10 — v8+ plans don't exist in production yet; the bug only triggers in a future-proofing scenario and user-authored data (queue_order/skipped/clusters/overrides) is preserved via setdefault(). |

## Scores

| Criterion | Score |
|-----------|-------|
| Significance | 5/10 |
| Originality | 4/10 |
| Core Impact | 2/10 |
| Overall | 3/10 |

## Summary

upgrade_plan_to_v7() unconditionally forces plan['version'] = V7_SCHEMA_VERSION at schema_migrations.py:304-305, meaning any plan with version > 7 is silently downgraded to v7 on the next save_plan() call. The core mechanism is confirmed: the missing version guard is a directional violation in migration code. The submission overstates impact ('silently erase user intent') since a stderr warning is printed and user data fields are preserved via setdefault(); only the version field and specific legacy keys are affected. Qualified YES_WITH_CAVEATS — real bug, overstated framing.

## Why Desloppify Missed This

- **What should catch:** A round-trip test: create a plan dict with version = V7_SCHEMA_VERSION + 1, call load_plan() (or directly ensure_plan_defaults()), then save_plan(), and assert the on-disk version is still > 7.
- **Why not caught:** All existing tests load plans at or below V7_SCHEMA_VERSION. The warning branch at persistence.py:59-65 was added for future-proofing but no test exercises the code path where a v8+ plan is loaded and then saved.
- **What could catch:** A parametric schema regression test covering version > PLAN_VERSION would catch this immediately. Alternatively, a property-based test that round-trips arbitrary version numbers through ensure_plan_defaults() and asserts version monotonicity.

## Verdict Files

- [Verdict JSON](https://github.com/xliry/desloppify/blob/fix/bounty-4000584288-agustif/bounty-verdicts/%40agustif-4000584288.json)
- [Verdict Report](https://github.com/xliry/desloppify/blob/fix/bounty-4000584288-agustif/bounty-verification-%40agustif-4000584288.md)

Generated with Lota

### DIFF
--- a/bounty-verdicts/@agustif-4000584288.json
+++ b/bounty-verdicts/@agustif-4000584288.json
@@ -4,6 +4,15 @@
   "verdict": "YES_WITH_CAVEATS",
   "verdict_date": "2026-03-06",
   "scoreboard_id": "S04",
+  "claim_restated": "load_plan() (persistence.py:67) calls ensure_plan_defaults() unconditionally for all plans. ensure_plan_defaults() (schema.py:198) always calls upgrade_plan_to_v7(), which at schema_migrations.py:304-305 sets plan['version'] = V7_SCHEMA_VERSION without checking whether the original version was already higher. A plan saved with version 8+ loaded by load_plan() has its in-memory version silently forced to 7; any subsequent save_plan() call (which also calls ensure_plan_defaults() at persistence.py:80) permanently writes version 7 to disk, downgrading the on-disk plan.",
+  "is_poor_engineering": true,
+  "why_poor_engineering": "A function named upgrade_plan_to_v7() should only act on plans with version < 7. The missing version guard — `if plan.get('version') < V7_SCHEMA_VERSION` — turns what should be a one-directional upgrade into an unconditional version clobber. Any plan that has advanced beyond v7 (e.g. after a schema bump) will be silently downgraded each time it passes through the load/save cycle. The function violates the basic contract of migration code: migrations should be directional and idempotent, not regressive.",
+  "is_significant": false,
+  "why_missed": {
+    "what_should_catch": "A round-trip test: create a plan dict with version = V7_SCHEMA_VERSION + 1, call load_plan() (or directly ensure_plan_defaults()), then save_plan(), and assert the on-disk version is still > 7.",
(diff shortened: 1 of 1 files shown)


xliry:
Closing — will reopen with single-agent flow

PR #278, opened 2026-03-07
feat: verify S032 @Midwest-AI-Solutions bounty — YES_WITH_CAVEATS

xliry:
**Submission:** https://github.com/peteromallet/desloppify/issues/204#issuecomment-4001579577
**Author:** @Midwest-AI-Solutions

## Problem (in our own words)

`_build_cluster_meta` in `plan_order.py:162` uses `str.replace(str(total_in_cluster), str(len(members)))` to update the issue count in cluster descriptions. This is a global substitution — it replaces every occurrence of the digit string, not just the count. If the old count appears as a substring of another number in the description, that number gets silently corrupted.

## Evidence

- `desloppify/engine/_work_queue/plan_order.py:159-162` at commit `6eb2065`: code matches submission exactly. `str.replace` with no `count` argument performs global substitution.
- `desloppify/engine/_plan/cluster_strategy.py:71-101` at `6eb2065`: `generate_description()` produces templates like `f"Fix {count} {label} issues"` — these contain only the count number, making collisions unlikely with current code.
- Bug is still present in HEAD (no changes to this function since snapshot).

## Fix

No fix applied — this is a verification-only PR. A proper fix would use `str.replace(old, new, 1)` for single replacement, or better yet, regenerate the description from structured data rather than patching the text.

## Verdict

| **Is this poor engineering?** | YES | `str.replace` for count substitution in natural language is a well-known anti-pattern |
| **Is this at least somewhat significant?** | MARGINAL | Current description templates embed only the count number, so digit collisions are unlikely in practice |

## Scores

| Criterion | Score |
|-----------|-------|
| Significance | 4/10 |
| Originality | 7/10 |
| Core Impact | 3/10 |
| Overall | 4/10 |

## Summary

The `str.replace` global substitution bug is correctly identified and well-explained. However, the dramatic examples ("112 files" → "88 files") don't match actual generated descriptions, which use simple templates containing only the count number. The practical corruption risk is low with current code, though it represents a valid latent fragility. Original finding — no duplicates.

## Why Desloppify Missed This

- **What should catch:** A code-quality detector for unsafe string substitution patterns (global replace on numeric substrings in natural language)
- **Why not caught:** No detector targets string manipulation anti-patterns; the code "works" for the common case
- **What could catch:** A pattern detector for `str.replace` with `str(int_var)` arguments, or a broader "fragile text manipulation" heuristic

## Verdict Files

- [Verdict JSON](https://github.com/xliry/desloppify/blob/task-468-lota-1/bounty-verdicts/%40Midwest-AI-Solutions-4001579577.json)
- [Verdict Report](https://github.com/xliry/desloppify/blob/task-468-lota-1/bounty-verification-%40Midwest-AI-Solutions-4001579577.md)

### DIFF
--- a/bounty-verdicts/@Midwest-AI-Solutions-4001579577.json
+++ b/bounty-verdicts/@Midwest-AI-Solutions-4001579577.json
@@ -1,0 +1,12 @@
+{
+  "submission_id": "S032",
+  "comment_id": 4001579577,
+  "author": "Midwest-AI-Solutions",
+  "title": "Naive str.replace in _build_cluster_meta corrupts cluster descriptions",
+  "verdict": "YES_WITH_CAVEATS",
+  "significance": 4,
+  "originality": 7,
+  "core_impact": 3,
+  "overall": 4,
+  "notes": "The str.replace global substitution bug is real and correctly identified at plan_order.py:162. However, practical impact is low: cluster descriptions are generated from templates (cluster_strategy.py:generate_description) that embed only the count number — no other digits to corrupt. The hypothetical examples (112→88) don't match actual generated descriptions. Still a valid code quality issue — str.replace is wrong for count patching."
+}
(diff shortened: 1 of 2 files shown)


peteromallet:

PR #311, opened 2026-03-07
feat: verify S154 @juzigu40-ui bounty — YES_WITH_CAVEATS, tri-state full-sweep logic (#500)

xliry:
**Submission:** https://github.com/peteromallet/desloppify/issues/204#issuecomment-4005589003
**Author:** @juzigu40-ui

## Problem (in our own words)

The `full_sweep_included` field uses tri-state logic (`True`/`False`/`None`) across the holistic review import chain, but `detect_review_coverage` checks `if full_sweep_included is not False` — treating both `True` and `None` as a full sweep. An empty holistic import (zero issues, no explicit `review_scope`) causes `full_sweep_included` to default to `None`, which is omitted from cache (only booleans stored). Subsequent scans read `None` from cache, treat it as full-sweep, suppress holistic coverage issue regeneration, and the previously auto-resolved markers escape all failure status sets — improving scores without actual review evidence.

## Evidence

- `holistic.py:62-66`: `review_scope.setdefault("full_sweep_included", None)` — defaults to `None` when absent
- `holistic_cache.py:97-98`: `if isinstance(full_sweep_included, bool)` — only stores explicit booleans, omits `None`
- `review_coverage.py:170-171`: `if full_sweep_included is not False` — `None is not False` → treated as full sweep
- `review_coverage.py:65`: `if holistic_fresh: return None` — suppresses unreviewed file detection
- `holistic_cache.py:124-133`: `resolve_holistic_coverage_issues` converts open holistic markers to `auto_resolved`
- `core.py:191-194`: `FAILURE_STATUSES_BY_MODE` excludes `auto_resolved` from all scoring modes

## Fix

Add explicit boolean checks: `if full_sweep_included is True` instead of `is not False`. Or ensure `update_holistic_review_cache` always records `full_sweep_included` (defaulting to `False` rather than omitting).

## Verdict

| **Is this poor engineering?** | YES | Tri-state `is not False` allows `None` to silently grant full-sweep privileges, creating a scoring integrity gap |
| **Is this at least somewhat significant?** | YES | Empty holistic imports can suppress coverage debt and inflate scores without review evidence |

## Scores

| Criterion | Score |
|-----------|-------|
| Significance | 5/10 |
| Originality | 5/10 |
| Core Impact | 4/10 |
| Overall | 5/10 |

## Summary

All 5 code references verified at snapshot 6eb2065. The tri-state `full_sweep_included` chain is real: `None` defaults flow from `import_holistic_issues` through `update_holistic_review_cache` (which omits non-booleans) to `detect_review_coverage` (where `None is not False` grants full-sweep status). This suppresses coverage issue regeneration while `auto_resolved` escapes all failure sets. Partial overlap with S118 (same downstream effect, different entry path). Requires specific conditions (empty holistic import without explicit scope) and may be intentional legacy handling.

## Why Desloppify Missed This

- **What should catch:** A tri-state / sentinel value detector that flags `is not False` / `is not None` patterns when the variable can legally hold 3+ states
- **Why not caught:** No detector tracks tri-state logic patterns or sentinel value confusion across function boundaries
- **What could catch:** A cross-function data-flow detector that traces default values through cache storage and retrieval, flagging asymmetric None handling

## Verdict Files

- [Verdict JSON](https://github.com/xliry/desloppify/blob/task-500-lota-1/bounty-verdicts/%40juzigu40-ui-4005589003.json)
- [Verdict Report](https://github.com/xliry/desloppify/blob/task-500-lota-1/bounty-verification-%40juzigu40-ui-4005589003.md)

### DIFF
--- a/bounty-verdicts/@juzigu40-ui-4005589003.json
+++ b/bounty-verdicts/@juzigu40-ui-4005589003.json
@@ -1,0 +1,12 @@
+{
+  "submission_id": "S154",
+  "comment_id": 4005589003,
+  "author": "juzigu40-ui",
+  "title": "Tri-state full-sweep logic allows evidence-free holistic attestation to suppress coverage debt",
+  "verdict": "YES_WITH_CAVEATS",
+  "significance": 5,
+  "originality": 5,
+  "core_impact": 4,
+  "overall": 5,
+  "notes": "All 5 code references verified. import_holistic_issues defaults full_sweep_included to None; update_holistic_review_cache only stores booleans so None is omitted; detect_review_coverage uses 'is not False' treating None as full sweep; resolve_holistic_coverage_issues auto-resolves markers; auto_resolved absent from FAILURE_STATUSES_BY_MODE. Real scoring integrity chain but requires empty holistic import without explicit review_scope. Partial overlap with S118 (same downstream None-as-full-sweep effect through different entry path)."
+}
--- a/bounty-verification-@juzigu40-ui-4005589003.md
+++ b/bounty-verification-@juzigu40-ui-4005589003.md
@@ -1,0 +1,65 @@
+# Bounty Verification: S154 @juzigu40-ui — Tri-State Full-Sweep Logic Suppresses Coverage Debt
+
+**Submission:** https://github.com/peteromallet/desloppify/issues/204#issuecomment-4005589003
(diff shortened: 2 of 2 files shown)


peteromallet:

PR #372, opened 2026-03-08
Fix prepared review packet reuse and runner hang regressions

### LINKED ISSUE #371: Review batch reruns can ignore prepared packets and hang runners
## Summary

`review --run-batches` can ignore the packet produced by `review --prepare`, which leads to stale prompt content, incorrect rerun gating, and unexpected full-dimension batch explosions. Separately, silent Codex runs can be killed as "stalled" before their real timeout, and `scan` can hang in the boilerplate duplication phase when `jscpd` is unavailable locally.

### 1. Prepared rerun scope is lost

1. Run:
   `desloppify review --prepare --path . --dimensions dependency_health`
2. Then run a dry batch execution from the prepared state:
   `desloppify review --run-batches --runner codex --dry-run --only-batches 1 --path .`

Expected:
- the prepared `dependency_health` packet is reused
- prompt 1 is still scoped to `dependency_health`
- rerun warnings/gating are based on that prepared scope

Actual:
- `run-batches` rebuilds a fresh holistic packet unless `--packet` or `--dimensions` are restated
- prompt content can drift to another dimension such as `cross_module_architecture`
- rerun warnings treat the run as though it targets the full default scored dimension set

### 2. Parallel batch execution can expand to all dimensions

Using the same prepared state as above:

`desloppify review --run-batches --runner codex --parallel --path .`

Expected:
- run only the prepared batch set

Actual:
- a fresh packet can be generated for the full default scored set, creating an unexpected 20-batch run

### 3. Silent Codex batches are killed early as stalled

When the Codex runner stays quiet before writing its result file, the stall detector can treat "missing output file + no stream activity" as a stall baseline and kill the batch before the configured timeout.

Expected:
- a batch with no output artifact yet should continue until the real timeout or actual process exit

Actual:
- the batch can be killed as "stalled" first

### 4. `scan` can hang at boilerplate duplication

Run:

`desloppify scan --path .`

in an environment where `jscpd` is not installed locally.

Expected:
- duplication analysis should skip quickly with a warning

Actual:
- the `npx --yes jscpd` path can hang while attempting package resolution / install behavior

## Root causes

- `run-batches` does not reuse `.desloppify/query.json` as the authoritative prepared holistic packet when no new scope is provided.
- rerun preflight only looks at explicit CLI dimensions instead of the prepared query scope.
- blind packet snapshots use a shared filename instead of a run-scoped filename.
- stall detection starts tracking stability even before an output file exists.
- `jscpd` execution prefers `npx --yes jscpd`, which is fragile in restricted or offline environments.

## Proposed fix

- reuse the prepared holistic query packet from `query.json` when `run-batches` follows `review --prepare` without explicit `--packet` / new `--dimensions`
- infer rerun target dimensions from the prepared query during preflight
- make blind packet snapshots run-scoped
- only allow stall detection after an output artifact exists
- prefer a local/global `jscpd` executable and skip quickly when it is unavailable

jaredkirby:
Fixes #371

## Summary

This fixes a group of review/scan workflow regressions that showed up while validating the Codex-based holistic review path:

- `review --run-batches` now reuses the prepared holistic packet from `.desloppify/query.json` when no new scope is requested
- rerun preflight now derives its target dimensions from the prepared query when appropriate
- blind packet snapshots are run-scoped instead of reusing a single global filename
- silent Codex batches are no longer killed as "stalled" before an output file exists
- `jscpd` duplication analysis now skips quickly when no local/global `jscpd` executable is available

## Why

These issues combined into a pretty confusing workflow:

- a prepared `dependency_health` rerun could generate a stale prompt for another dimension
- `run-batches --parallel` could unexpectedly expand into a full 20-batch run
- every silent batch could then fail with early stall kills
- `scan` could hang at boilerplate duplication in environments without `jscpd`

## Validation

Targeted tests:

- `pytest desloppify/tests/commands/review/test_review_preflight.py desloppify/tests/review/test_runner_internals.py desloppify/tests/detectors/test_external_adapters.py desloppify/tests/review/review_commands_cases.py -q`

Result:

- `249 passed`

## Notes

The prepared-packet reuse path intentionally stays conservative:

- explicit `--packet` still wins
- explicit `--dimensions` still triggers fresh packet preparation
- retrospective runs still prepare a fresh packet

### DIFF
--- a/desloppify/app/commands/helpers/query.py
+++ b/desloppify/app/commands/helpers/query.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import json
 import sys
 from dataclasses import dataclass
 from pathlib import Path
@@ -40,6 +41,16 @@ def query_writer() -> QueryWriter:
     return QueryWriter(query_file=query_file_path())
 
 
+def load_query() -> dict | None:
+    """Load the active query payload, returning ``None`` on read/parse failure."""
+    path = query_file_path()
+    try:
+        payload = json.loads(path.read_text())
+    except (OSError, json.JSONDecodeError):
+        return None
+    return payload if isinstance(payload, dict) else None
+
+
 def write_query(data: dict) -> OutputResult:
     """Write structured query output using default best-effort policy."""
     return write_query_best_effort(data, context="query payload update")
@@ -52,6 +63,7 @@ def write_query_best_effort(data: dict, *, context: str) -> OutputResult:
 
 __all__ = [
     "QueryWriter",
+    "load_query",
     "query_file_path",
     "query_writer",
     "write_query",
--- a/desloppify/app/commands/review/_runner_process_io.py
+++ b/desloppify/app/commands/review/_runner_process_io.py
@@ -187,12 +187,10 @@ def _check_stall(
     except OSError:
         current_signature = None
     if current_signature is None:
-        baseline = prev_stable if isinstance(prev_stable, int | float) else now
-        output_age = now - baseline
-        stream_idle = now - last_activity
-        if output_age >= threshold and stream_idle >= threshold:
-            return True, None, baseline
-        return False, None, baseline
+        # Codex may remain silent until it writes the final output artifact.
+        # Only treat a batch as stalled once an output file exists and then
+        # remains stable without stream activity.
+        return False, None, None
     if current_signature != prev_sig:
         return False, current_signature, now
     if prev_stable is None:
--- a/desloppify/app/commands/review/batch/orchestrator.py
+++ b/desloppify/app/commands/review/batch/orchestrator.py
@@ -8,6 +8,7 @@ import sys
 from pathlib import Path
 from typing import cast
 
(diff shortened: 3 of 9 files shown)


peteromallet:
Merged manually with rebase to resolve conflicts from the batch/core.py split — thanks @jaredkirby for the thorough regression fixes! Author credit preserved in commit.

Of the 5 original fixes, 3 had already been resolved on main during the `batch/core.py` refactor:
- **Stall detection for silent Codex batches** — already fixed; `_check_stall` no longer declares stalls when no output file exists
- **Run-scoped blind packet paths** — addressed in `runtime_paths.py` (stamp-based paths)
- **jscpd skip when unavailable** — already implemented via `_resolve_jscpd_command()`

The 2 fixes applied:
- **Preflight dimension resolution** — `_target_dimensions()` now resolves dimensions from `query.json` for `run-batches`/`external-start` modes
- **`load_query()` helper** — shared utility for reading prepared query payloads

Commit: 96786dd

PR #391, opened 2026-03-11
fix: improve state locking mechanism for cross-platform compatibility

mortael:
The same as https://github.com/peteromallet/desloppify/commit/f954dff86616029eb362c95f8d958a2edcc1c0eb, but now for _state


peteromallet:
Implemented in 0.9.6 (67b8f3f9) with credit. Thanks @mortael!

PR #404, opened 2026-03-12
fix: rust workspace rustdoc execution

MacHatter1:
## Context

This is a follow-up fix for a regression introduced in the earlier Rust support PR.

That PR added the Rust `rustdoc_warning` phase, but the command it used was:

```bash
cargo rustdoc --workspace ...
```

`cargo rustdoc` does not accept `--workspace`, so for Cargo workspaces the rustdoc phase never actually ran. The detector degraded, but the intended rustdoc validation was missing.

## What this PR fixes

This PR fixes that regression by keeping the phase on real `cargo rustdoc`, but running it correctly for workspaces:

1. Run `cargo metadata --format-version=1 --no-deps`
2. Find workspace members that expose a library target
3. Run `cargo rustdoc -p <package> --lib ...` for each eligible package
4. Merge rustdoc diagnostics across packages
5. Keep hard failures visible when a package rustdoc run fails without parseable diagnostics

This is a correction to the Rust PR’s execution model, not a new Rust feature.

## Why this fix

- It fixes the actual bug introduced by the Rust PR
- It keeps the semantics correct by using `cargo rustdoc`, not `cargo doc`
- It works for multi-crate workspaces
- It skips binary-only packages that do not have a library rustdoc target
- It does not hide real failures

## Files changed

- `desloppify/languages/rust/tools.py`
  - added workspace package discovery and per-package rustdoc execution
- `desloppify/languages/rust/commands.py`
  - routed the direct detector through the new runner
- `desloppify/languages/rust/phases.py`
  - routed the scan phase through the same runner
- `desloppify/languages/rust/tests/test_tools.py`
  - added regression tests for package selection, per-package execution, merged diagnostics, and failure handling
- `desloppify/languages/rust/README.md`
  - updated docs to describe the real rustdoc policy

## Validation

Targeted Rust tests:

- `python3 -m pytest desloppify/languages/rust/tests/test_tools.py desloppify/languages/rust/tests/test_init.py`
- Result: `12 passed`

Command-layer + Rust coverage:

- `python3 -m pytest desloppify/tests/commands desloppify/languages/rust/tests`
- Result: `1588 passed`

Full repo regression run:

- `python3 -m pytest`
- Result: `5866 passed, 145 skipped`

Open-source workspace verification:

- Verified against the open-source workspace example `Miosso/rust-workspace`
- Ran:

```bash
python3 -m desloppify.cli --lang rust detect rustdoc_warning --path <workspace-root> --json
```

- Result: returned a real rustdoc finding from the workspace library crate and skipped the binary-only members

- Also ran:

```bash
python3 -m desloppify.cli --lang rust scan --path <workspace-root> --force-rescan --attest "I understand this is not the intended workflow and I am intentionally skipping queue completion"
```

- Result: completed successfully and reached the `cargo rustdoc` phase inside the normal scan flow

## Reviewer focus

The main thing to review is whether the per-package `cargo rustdoc` runner is the right correction for the regression introduced by the Rust PR, and whether the failure semantics remain appropriately strict.

### DIFF
--- a/desloppify/languages/rust/README.md
+++ b/desloppify/languages/rust/README.md
@@ -308,7 +308,7 @@ Current command policy:
 
 - Clippy runs workspace-wide, all targets, all features, JSON output
 - Cargo check runs workspace-wide, all targets, all features, JSON output
-- Rustdoc runs workspace-wide, all features, library target only, JSON output
+- Rustdoc runs once per workspace library package with `cargo rustdoc -p <package> --lib`, all features, JSON output
 
 Current rustdoc warnings enabled:
 
--- a/desloppify/languages/rust/commands.py
+++ b/desloppify/languages/rust/commands.py
@@ -23,6 +23,7 @@ from desloppify.languages._framework.commands_base_registry import (
     make_cmd_dupes,
     make_cmd_orphaned,
 )
+from desloppify.languages._framework.generic_parts.tool_runner import ToolRunResult
 from desloppify.languages._framework.generic_parts.tool_runner import run_tool_result
 from desloppify.languages.rust.detectors import (
     detect_async_locking,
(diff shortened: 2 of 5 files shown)


peteromallet:
Cherry-picked into `desloppify/code-health-20260312` at 52c32a04 — will be included in 0.9.6. Solid fix @MacHatter1, the per-package approach is the right call and the test coverage is thorough.

PR #413, opened 2026-03-13
Fix planner deadlock when triage goes stale mid-cycle

imetandy:
## Summary
I ran into a planner deadlock after a trusted review import landed mid-cycle. New review issues made triage stale, but I still had open objective work in the queue. In that state, `desloppify plan resolve ... --confirm` was blocked because triage was stale, while `desloppify plan triage --run-stages --runner codex` was also blocked because objective backlog was still open.

This change makes that state consistent instead of deadlocking it. When triage is stale but still deferred behind objective backlog, objective resolves can continue, direct review resolves stay blocked, and the CLI now says `TRIAGE PENDING` instead of nudging me toward a triage command that cannot run yet.

## What I changed
- detect when stale triage is pending behind objective backlog
- allow objective `fixed` resolves to proceed in that pending state
- keep direct review resolves blocked until objective backlog is cleared
- align banner and lifecycle copy around `TRIAGE PENDING`
- pass the already-loaded plan into the resolve guard instead of reloading it
- add regression coverage for the pending-vs-blocked behavior

## Testing
```bash
pytest desloppify/tests/commands/helpers/test_guardrails.py desloppify/tests/plan/test_triage_phase_banner.py desloppify/tests/commands/resolve/test_cmd_resolve.py desloppify/tests/plan/test_triage_snapshot.py desloppify/tests/commands/plan/test_workflow_gates.py desloppify/tests/commands/plan/test_triage_dependency_guard.py
```

### DIFF
--- a/desloppify/app/commands/helpers/guardrails.py
+++ b/desloppify/app/commands/helpers/guardrails.py
@@ -5,9 +5,11 @@ from __future__ import annotations
 import logging
 from dataclasses import dataclass, field
 
+from desloppify import state as state_mod
 from desloppify.app.commands.helpers.issue_id_display import short_issue_id
 from desloppify.base.exception_sets import PLAN_LOAD_EXCEPTIONS, CommandError
 from desloppify.base.output.terminal import colorize
+from desloppify.engine._plan.sync.context import has_objective_backlog, is_mid_cycle
 from desloppify.engine.plan_state import load_plan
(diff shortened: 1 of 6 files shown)


peteromallet:
The fix for this is already on the 0.9.9 branch (c08b84de and subsequent commits built on top). Thanks for the clean PR and tests @imetandy — shipping in 0.9.9.

PR #424, opened 2026-03-14
fix(r): correct shell quote escaping in lintr command

sims1253:
Bug: The R language plugin's lintr command had broken shell quote escaping that prevented it from executing properly.

   Root Cause:
   1. Quote collision: The outer Python string used single quotes, but the shell command also used single quotes for the R expression, causing the shell to receive malformed commands.
   2. Over-engineering: The cat(paste(capture.output(...), collapse="\\n")) wrapper was unnecessary — lintr::lint_dir() already prints to stdout by default.

   The Fix:

```python
     # Before (broken):
     'cmd': (
         'Rscript -e \'cat(paste(capture.output('
         'lintr::lint_dir(".", show_notifications=FALSE)'
         '), collapse="\\n"))\''
     ),

     # After (fixed):
     "cmd": (
         "Rscript -e \"lintr::lint_dir('.')\""
     ),
```
   Changes:
   •  Switched to double quotes for the Python string, allowing single quotes inside for the R string
   •  Removed unnecessary cat/paste/capture.output wrapper
   •  Simplified to rely on lintr's default stdout behavior
   •  Output remains GNU-style formatted lint messages (compatible with fmt: "gnu")

### DIFF
--- a/desloppify/languages/r/__init__.py
+++ b/desloppify/languages/r/__init__.py
@@ -10,9 +10,7 @@ generic_lang(
         {
             "label": "lintr",
             "cmd": (
-                'Rscript -e \'cat(paste(capture.output('
-                'lintr::lint_dir(".", show_notifications=FALSE)'
-                '), collapse="\\n"))\''
+                "Rscript -e \"lintr::lint_dir('.')\""
             ),
             "fmt": "gnu",
             "id": "lintr_lint",


peteromallet:
Thanks @sims1253! Clean fix — the nested single-quote escaping was indeed broken for `shlex.split`. We simplified to double-quoted Python strings which resolves it properly.

Merged into `0.9.10` as `0f41dca3` with your attribution. Appreciate the contribution!

PR #425, opened 2026-03-14
feat(r): add Jarl as fast R linter with autofix

sims1253:
Summary
   Added Jarl (Just Another R Linter) as the fast, modern R linter with
    auto-fix support.

   Changes
   •  Added Jarl as tier 2 linter (becomes primary)
   •  Demoted lintr to tier 3 (fallback for projects without Jarl)
   •  Updated module docstring

   Why Jarl?
   •  Speed: Orders of magnitude faster than lintr (0.131s vs 18.5s on dplyr)
   •  Auto-fix: Supports --fix flag for automatic fixes (lintr has no auto-fix)
   •  Active development: Built on Air (Posit's Rust formatter)
   •  55+ rules: Growing rule set inspired by lintr

   Testing
   •  All 416 tests pass
   •  R import resolver test passes

### DIFF
--- a/desloppify/languages/r/__init__.py
+++ b/desloppify/languages/r/__init__.py
@@ -1,4 +1,4 @@
-"""R language plugin — lintr + tree-sitter."""
+"""R language plugin — Jarl, lintr + tree-sitter."""
 
 from desloppify.languages._framework.generic_support.core import generic_lang
 from desloppify.languages._framework.treesitter import R_SPEC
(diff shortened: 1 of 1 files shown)


peteromallet:
Thanks @sims1253! Jarl at tier-2 with autofix makes a lot of sense — 140x faster than lintr is a no-brainer, and `--allow-dirty` is the right call for a dev tool. 

Merged into `0.9.10` as `55f9aaf0` with your attribution. Nice contribution!

PR #429, opened 2026-03-15
Fix Rust dep graph hangs by ignoring string-literal fake imports and indexing path matches

fluffypony:
### Summary
This PR fixes a Rust dep-graph performance/pathology issue that could cause coupling analysis to stall or run very slowly on medium-sized codebases with complex `use` trees. On one of my codebases it ran up to 7.8gb of RAM, and ran for about 3 hours, before I killed it (M4 Max, so it's not a machine performance issue).

### Problem
Rust import extraction used a regex over comment-stripped source and could accidentally match `use ...;` text inside multiline string literals (for example large JSON/tool description strings).  
That produced giant pseudo-import specs and expensive recursive expansion work.  
On top of that, path candidate matching repeatedly scanned the full production file set in hot loops.

### Root Cause
1. `iter_use_specs` / `iter_pub_use_specs` matched imports in string content.
2. `_candidate_matches` did repeated O(N) scans in path resolution.
3. Alias/workspace metadata was repeatedly rebuilt in frequent resolution paths.

### Changes
1. In `desloppify/languages/rust/support.py`, import extraction now masks string literal content (while preserving line/offset alignment) before running import regexes, then expands matches using the original stripped text slice.
2. Added `RustProductionFileIndex` + `build_production_file_index` for absolute/relative O(1) lookup-style matching.
3. Threaded optional `production_index` through Rust resolution helpers (`resolve_use_spec`, `resolve_mod_declaration`, `resolve_barrel_targets`, `_resolve_from_source_root`, `_match_root_files`, `_candidate_matches`).
4. Updated `desloppify/languages/rust/detectors/deps.py` to build one production index per graph build and pass it through.
5. Added caching for local dependency alias index building.
6. Added regressions in `desloppify/languages/rust/tests/test_support.py`:
1. `test_iter_use_specs_ignores_use_text_inside_strings`
2. `test_match_production_candidate_uses_relative_index_key`

### Validation
1. `pytest -q desloppify/languages/rust/tests/test_support.py desloppify/languages/rust/tests/test_deps.py desloppify/languages/rust/tests/test_test_coverage.py` → `22 passed`
2. Repro against `~/Desktop/working/nsh`:
1. `build_dep_graph` (mod edges): ~5.54s
2. `build_dep_graph` (no mod edges): ~2.87s

### Impact
1. Eliminates the observed dep-graph hang/near-hang pattern from string-literal false-positive imports.
2. Reduces path resolution overhead in Rust dep graph construction.
3. Preserves existing dep-graph semantics while improving robustness and throughput.

### DIFF
--- a/desloppify/languages/rust/detectors/deps.py
+++ b/desloppify/languages/rust/detectors/deps.py
@@ -7,6 +7,7 @@ from typing import Any
 
 from desloppify.engine.detectors.graph import finalize_graph
 from desloppify.languages.rust.support import (
+    build_production_file_index,
     build_workspace_package_index,
     find_rust_files,
     iter_mod_targets,
@@ -31,6 +32,7 @@ def build_dep_graph(
         return {}
 
     file_set = set(graph.keys())
+    production_index = build_production_file_index(file_set)
     package_index = build_workspace_package_index()
     for filepath in files:
         content = read_text_or_none(filepath)
@@ -44,6 +46,7 @@ def build_dep_graph(
                     filepath,
                     file_set,
                     declared_path=declared_path,
+                    production_index=production_index,
                 )
                 if resolved and resolved != filepath:
                     graph[filepath]["imports"].add(resolved)
@@ -56,6 +59,7 @@ def build_dep_graph(
                 file_set,
                 package_index,
                 allow_crate_root_fallback=False,
+                production_index=production_index,
             )
             if resolved and resolved != filepath:
                 graph[filepath]["imports"].add(resolved)
--- a/desloppify/languages/rust/support.py
+++ b/desloppify/languages/rust/support.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import functools
 import re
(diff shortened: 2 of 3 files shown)


peteromallet:
Thanks @fluffypony! Excellent root cause analysis on the string-literal false imports causing exponential expansion. The fix is well-designed — masking before regex matching, O(1) file indexing, and LRU caching for workspace/alias rebuilds. 

Minor note for future-proofing: Rust 1.77+ C-string literals (`c"..."` / `cr#"..."#`) aren't masked yet, but that's low risk since they're unlikely to contain `use` statements. Can address if it ever comes up.

Merged into `0.9.10` as `ba9f1907` with your attribution.

fluffypony:
Awesome thanks so much!

peteromallet:
This was landed directly on the 0.9.10 branch (commit ba9f1907). Closing the PR but the changes are in — thanks for tracking down the Rust dep graph hang, @fluffypony!

PR #439, opened 2026-03-15
fix: project root detection, force-rescan plan wipe, and auto-cluster triage

peteromallet:
## Summary

Three fixes discovered while running desloppify on a real project:

- **fix: derive project root from state_file path in do_import_run follow-up scan** — `--scan-after-import` was writing state to `.desloppify/subagents/runs/<stamp>/results/.desloppify/` instead of the project root's `.desloppify/`. Root cause: `_runtime_project_root()` returned the wrong path. Fix: use `Path(state_file).parent.parent` which is always correct.

- **fix: preserve plan_start_scores during force-rescan to protect manual clusters** — `--force-rescan` was wiping manual cluster items while keeping cluster names. Root cause: `_reset_cycle_for_force_rescan()` cleared `plan_start_scores`, making `is_mid_cycle()` return False, which triggered full cluster regeneration and issue ID reconciliation that superseded all old IDs. Fix: don't clear `plan_start_scores` during force-rescan.

- **feat: require explicit triage decisions for auto-clusters** — Auto-clusters (auto/unused, auto/security, etc.) entered the execution queue but were invisible to triage ("silence means leave in backlog"). Now the triage prompt requires an explicit decision (promote/skip/break_up) for each auto-cluster, with a structured `auto_cluster_decisions` field in the triage output schema.

## Test plan

- [x] All 5380 existing tests pass
- [ ] Run `desloppify review --import-run <dir> --scan-after-import` and verify state writes to correct location
- [ ] Run `desloppify scan --force-rescan` with manual clusters and verify they survive
- [ ] Run triage and verify auto-clusters appear with "decision required" prompt

### DIFF
--- a/RELEASE_NOTES_0.9.10.md
+++ b/RELEASE_NOTES_0.9.10.md
@@ -6,7 +6,7 @@ This release adds **experimental Hermes Agent integration** for fully autonomous
 
 ---
 
-**124 files changed | 26 commits | 5,425 tests passing**
+**141 files changed | 47 commits | 5,438 tests passing**
 
 ## Hermes Agent Integration (Experimental)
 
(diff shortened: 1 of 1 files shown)


peteromallet:
This was landed directly on the 0.9.10 branch (commit 0cbd123e). Closing the PR — changes are in.

PR #440, opened 2026-03-16
Add Rust inline-test filtering

fluffypony:
## Summary
This PR makes Rust clippy ingestion inline-test-aware so diagnostics from `#[cfg(test)] mod ...` blocks in `src/*.rs` are not counted as production findings.

## Problem
`cargo clippy --message-format=json` reports spans by file/line, and inline unit tests live in production files. Desloppify previously treated those diagnostics as production issues, which inflated strict-score debt (especially `clippy::unwrap_used` / `clippy::expect_used`) even when findings were test-only.

## What Changed
1. Added clippy-specific filtering in [`parse_clippy_messages`](desloppify/languages/rust/tools.py) to optionally skip diagnostics inside inline test modules.
2. Added source-range detection for inline test modules guarded by cfg predicates that require `test` (for example `cfg(test)` and `cfg(all(test, ...))`).
3. Implemented syntax-aware scanning helpers so delimiter/attribute parsing remains stable around comments and literals:
- line/block comments
- strings and raw strings
- char literals (without misclassifying lifetimes)
- nested attribute brackets
4. Extended inline module matching to support raw identifiers (for example `mod r#tests { ... }`).
5. Kept cfg logic conservative: ambiguous/non-test forms (for example `cfg(any(test, ...))` or `cfg(not(test))`) are not auto-skipped.

## Test Coverage
Added/expanded parser tests in [`desloppify/languages/rust/tests/test_tools.py`](desloppify/languages/rust/tests/test_tools.py), including:
1. skip diagnostics in `cfg(test)` inline modules
2. preserve non-test diagnostics in the same file
3. keep `cfg(not(test))`
4. skip `cfg(all(test, ...))`
5. keep `cfg(any(test, ...))`
6. handle comments/doc comments between cfg attr and module
7. ignore commented-out cfg markers
8. handle braces/URLs/lifetimes in test code without breaking module range detection
9. support raw identifier module names
10. handle intervening attrs like `#[doc = "]"]`

## Validation
```bash
pytest -q desloppify/languages/rust/tests/test_tools.py
pytest -q desloppify/languages/rust/tests/test_custom.py
ruff check desloppify/languages/rust/tools.py desloppify/languages/rust/tests/test_tools.py
```

All checks passed.

## Impact
This change removes test-only clippy noise from production scoring while preserving existing behavior for non-test diagnostics and other Rust tool parsers (`cargo check`, `rustdoc`).

### DIFF
--- a/desloppify/languages/rust/tests/test_tools.py
+++ b/desloppify/languages/rust/tests/test_tools.py
@@ -49,6 +49,532 @@ def test_parse_clippy_messages_ignores_non_json_noise():
     ]
 
 
+def test_parse_clippy_messages_skips_inline_cfg_test_module_diagnostics(tmp_path):
+    source = tmp_path / "src" / "lib.rs"
+    source.parent.mkdir(parents=True, exist_ok=True)
+    source.write_text(
+        """
+pub fn runtime_value() -> usize {
+    1
+}
+
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn inline_test_uses_unwrap() {
+        let _ = Some(1usize).unwrap();
+    }
+}
+""".strip()
+        + "\n"
+    )
+
+    message = {
+        "reason": "compiler-message",
+        "message": {
+            "level": "warning",
+            "message": "used `unwrap()` on an `Option` value",
+            "code": {"code": "clippy::unwrap_used"},
+            "spans": [
+                {
+                    "is_primary": True,
+                    "file_name": "src/lib.rs",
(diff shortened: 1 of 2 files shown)


peteromallet:
This is really well done — the conservative approach to cfg predicate parsing is exactly right, and the test coverage is thorough. The hand-rolled syntax scanner for brace matching is substantial but the edge case tests give me confidence. Merging, thanks @fluffypony!

PR #448, opened 2026-03-16
fix: wire communicate-score cycle-boundary sentinels

### LINKED ISSUE #447: Fix communicate-score cycle-boundary sentinel resets
## Summary

`workflow::communicate-score` can re-inject incorrectly around cycle transitions because the resolved-this-cycle guard is not fully wired through the framework's cycle-boundary reset points.

## Repro

1. Let `workflow::communicate-score` inject and resolve it.
2. Trigger a follow-up scan / reconcile pass.
3. The workflow gate can re-evaluate after `previous_plan_start_scores` is cleared.
4. In objective-only or PRR-style flows, `sync_communicate_score_needed()` may re-inject unexpectedly unless a separate resolved-this-cycle sentinel survives the same cycle.

## Root cause

The guard logic needs two pieces:

- a `communicate_score_resolved_this_cycle` sentinel checked by `sync_communicate_score_needed()`
- a boundary reset helper that clears that sentinel only at true cycle boundaries

Without both halves wired consistently, one of two bad outcomes happens:

- the item re-injects after resolve + rescan in the same cycle, or
- the sentinel survives too long and suppresses the next real cycle

## Proposed fix

- gate `sync_communicate_score_needed()` on `communicate_score_resolved_this_cycle`
- set the sentinel when the workflow item is injected
- clear it at the same true cycle boundaries as the other workflow sentinels:
  - force-rescan reset
  - plan-start score seeding
  - queue-drain / plan-start clear
  - trusted review import

## Regression coverage

Add focused tests for:

- helper-level clearing of `clear_communicate_score_sentinel()`
- score seeding clearing all workflow sentinels
- force-rescan clearing all workflow sentinels
- queue-drain clearing all workflow sentinels
- trusted import clearing all workflow sentinels

cdunda-perchwell:
## Summary
- add a `communicate_score_resolved_this_cycle` workflow guard so `workflow::communicate-score` does not re-inject after resolve + rescan in the same cycle
- clear that sentinel at the same true cycle boundaries as the sibling workflow sentinels
- add focused regressions for scan reconcile, trusted import, and helper-level sentinel clearing

## Problem
`workflow::communicate-score` had a resolved-this-cycle sentinel pattern, but the boundary reset wiring was incomplete. That meant the framework could either:
- re-inject the workflow unexpectedly after resolve + rescan in the same cycle, or
- suppress it too long if the new sentinel was not cleared at the next real cycle boundary

## Fix
- gate `sync_communicate_score_needed()` on `communicate_score_resolved_this_cycle`
- set the sentinel at injection time
- clear it at:
  - force-rescan reset
  - plan-start score seeding
  - queue-drain / plan-start clear
  - trusted review import

## Tests
- `uv run python -m pytest -q desloppify/tests/plan/test_reconcile_pipeline.py -k 'clear_communicate_score_sentinel or clear_score_communicated_sentinel or communicate_score'`
- `uv run python -m pytest -q desloppify/tests/commands/scan/test_plan_reconcile.py desloppify/tests/commands/scan/test_plan_reconcile_postflight_and_reconcile.py desloppify/tests/commands/review/test_review_importing_support_direct.py desloppify/tests/commands/test_lifecycle_transitions.py desloppify/tests/plan/test_reconcile_pipeline.py -k 'communicate or sentinel or plan_start_scores or force_rescan or trusted_internal_clears'`

Fixes #447.

### DIFF
--- a/desloppify/app/commands/review/importing/plan_sync.py
+++ b/desloppify/app/commands/review/importing/plan_sync.py
@@ -29,6 +29,7 @@ from desloppify.engine._plan.sync import (
 )
 from desloppify.engine._plan.sync.workflow_gates import sync_import_scores_needed
 from desloppify.engine._plan.sync.workflow import (
+    clear_communicate_score_sentinel,
     clear_create_plan_sentinel,
     clear_score_communicated_sentinel,
 )
@@ -251,6 +252,7 @@ def _apply_import_plan_transitions(
     if trusted:
         clear_score_communicated_sentinel(plan)
         clear_create_plan_sentinel(plan)
+        clear_communicate_score_sentinel(plan)
         if sync_inputs.covered_ids:
             mark_subjective_review_completed(
                 plan,
--- a/desloppify/app/commands/scan/plan_reconcile.py
+++ b/desloppify/app/commands/scan/plan_reconcile.py
@@ -29,6 +29,7 @@ from desloppify.engine._plan.sync import (
 from desloppify.engine._plan.sync.dimensions import current_unscored_ids
 from desloppify.engine._plan.sync.context import is_mid_cycle
 from desloppify.engine._plan.sync.workflow import (
+    clear_communicate_score_sentinel,
     clear_create_plan_sentinel,
     clear_score_communicated_sentinel,
 )
@@ -52,6 +53,8 @@ def _reset_cycle_for_force_rescan(plan: dict[str, object]) -> bool:
         order.remove(item)
     clear_score_communicated_sentinel(plan)
     clear_create_plan_sentinel(plan)
+    clear_communicate_score_sentinel(plan)
+    plan.pop("scan_count_at_plan_start", None)
     meta = plan.get("epic_triage_meta", {})
     if isinstance(meta, dict):
         meta.pop("triage_recommended", None)
@@ -108,6 +111,7 @@ def _seed_plan_start_scores(plan: dict[str, object], state: state_mod.StateModel
     }
     clear_score_communicated_sentinel(plan)
     clear_create_plan_sentinel(plan)
+    clear_communicate_score_sentinel(plan)
     plan["scan_count_at_plan_start"] = int(state.get("scan_count", 0) or 0)
     return True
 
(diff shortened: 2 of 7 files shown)


cdunda-perchwell:
## Status update

This PR is **incomplete** — it partially addresses the issue but does not fully fix it.

### What is in this PR

- `communicate_score_resolved_this_cycle` sentinel added to `sync_communicate_score_needed`
- `clear_communicate_score_sentinel()` helper wired at force-rescan, score seeding, queue drain, trusted import

### What is missing

**Path 1 fix (scan path):** `_clear_plan_start_scores_if_queue_empty` in `plan_reconcile.py` still calls `clear_communicate_score_sentinel(plan)` when the queue drains. This clears the sentinel too early — on the next scan, both guards (`previous_plan_start_scores` and `communicate_score_resolved_this_cycle`) are gone and `sync_communicate_score_needed` re-injects.

The fix is one line: remove `clear_communicate_score_sentinel(plan)` from `_clear_plan_start_scores_if_queue_empty`. The sentinel should survive until `_seed_plan_start_scores` clears it at the start of the next cycle.

**Path 2 (resolve path):** Re-injection also occurs immediately during the resolve itself (same timestamp as the `done` log entry), before any rescan. Root cause not yet confirmed — simulation does not reproduce it. See issue #447 for full details.

### Tests

The existing test `test_clears_and_copies_to_state` was asserting the old (buggy) behavior (`communicate_score_resolved_this_cycle not in plan`). It needs to be updated to assert the sentinel IS preserved after queue drain.

A new regression test `test_communicate_score_not_reinjected_after_resolve` covering the two-scan scenario is written and passing locally but not yet pushed.

peteromallet:
Hey, thanks for the sentinel fix — the cycle-boundary logic looks solid and I want to get it merged. However this PR also bundles the entire `prr` plugin (~19k lines), resolve hooks, scoring changes, and broader exception handling that aren't mentioned in the title. Could you split this into separate PRs? Happy to merge the sentinel fix immediately once it's isolated.

The `prr` plugin and resolve hooks framework deserve their own focused reviews. Mixing them in here makes it really hard to review either piece properly.

cdunda-perchwell:
derp! Should be better now! @peteromallet

peteromallet:
Fix cherry-picked into v0.9.10 — communicate-score sentinel resets are resolved. Thank you @cdunda-perchwell for identifying both the original issue (#447) and the sentinel reset bug!

PR #450, opened 2026-03-16
feat(r): add R-specific code smell detectors

sims1253:
## Summary

 Adds 10 R-specific code smell checks to detect common R anti-patterns:

 | Smell | Description |
 |-------|-------------|
 | `setwd()` | Changes global working directory |
 | `<<-` | Global assignment operator |
 | `attach()` | Attaches data to search path |
 | `rm(list=ls())` | Clears entire workspace |
 | `browser()`/`debug()` | Debug leftovers |
 | `T`/`F` | Ambiguous boolean shortcuts |
 | `1:n()` | Off-by-one risk with empty vectors |
 | `options(stringsAsFactors)` | Deprecated pattern |
 | `library()` in functions | Side effects in functions |

 Also extends the generic language framework with `custom_phases` support, allowing any generic
 plugin to inject language-specific detector phases without converting to a full plugin.

 ## Files changed

 - `desloppify/languages/_framework/generic_support/core.py` (+5 lines)
 - `desloppify/languages/_framework/generic_support/registration.py` (+5 lines)
 - `desloppify/languages/r/__init__.py` (+7 lines)
 - `desloppify/languages/r/detectors/__init__.py` (new)
 - `desloppify/languages/r/detectors/smells.py` (+150 lines)
 - `desloppify/languages/r/detectors/smells_catalog.py` (+70 lines)
 - `desloppify/languages/r/phases_smells.py` (+25 lines)
 - `desloppify/languages/r/tests/__init__.py` (new)
 - `desloppify/languages/r/tests/test_r_smells.py` (+161 lines)

 ## Testing

 Includes comprehensive test coverage for all 10 smell detectors.

 ## Dependencies

 None - this is a foundational PR that other R enhancements can build on.

### DIFF
--- a/desloppify/languages/r/detectors/smells.py
+++ b/desloppify/languages/r/detectors/smells.py
@@ -2,11 +2,14 @@
 
 from __future__ import annotations
 
+import logging
 import re
 from pathlib import Path
 
 from desloppify.base.discovery.source import find_source_files
 from .smells_catalog import R_SMELL_CHECKS, SEVERITY_ORDER
+
+logger = logging.getLogger(__name__)
 
 _R_STRING_RE = re.compile(r'"(?:[^"\\]|\\.)*"|\'(?:[^\'\\]|\\.)*\'')
 _R_COMMENT_RE = re.compile(r"#[^\n]*")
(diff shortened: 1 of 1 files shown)


peteromallet:
Really like the detector coverage here — the 10 detectors are well-chosen and the test suite is solid for the happy paths. A few things to fix before I can merge:

1. **`_detect_library_in_function` brace tracking** — the current logic increments `fn_depth` for every `{` on any line, not just function definitions. This means any bare `{}` block (if/for/while) outside a function will set fn_depth > 0 and cause false positives for top-level `library()` calls that follow. Also `opens` appears to be computed but unused. Could you rework this to only track function-scoped braces?

2. **`_strip_r_comments` doesn't handle strings** — `#` inside string literals (e.g. `x <- "value #1"`) gets stripped. The docstring says 'preserving string literals' but the regex `#[^\n]*` doesn't account for them.

3. **File discovery bypasses framework** — `_find_r_files` rolls its own discovery with hardcoded exclusions instead of using `base/source_discovery.py`. This means it won't respect project-configured exclusion patterns. Would be great to use the framework's discovery instead.

Happy to help if you want to pair on any of these. The detectors themselves are exactly what R support needs.

sims1253:
I'll take a look tomorrow.

PR #428, opened 2026-03-14
feat: add SCSS language plugin

klausagnoletti:
## Summary

- Adds a new SCSS/Sass language plugin using the `generic_lang` framework
- Integrates stylelint with JSON output parsing for linting
- Supports `.scss` and `.sass` file extensions with `node_modules`, `_output`, `.quarto`, and `vendor` excluded
- Detects projects via `_scss` and `.stylelintrc` markers

## Test plan

- [ ] Install stylelint globally: `npm install -g stylelint stylelint-config-standard-scss`
- [ ] Run plugin tests: `pytest desloppify/languages/scss/tests/ -v`
- [ ] Run a scan against a SCSS project: `desloppify scan --path <scss-project>`
- [ ] Verify stylelint issues are surfaced correctly in output

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #452, opened 2026-03-16
feat(scss): tests, bug fixes, and truthful docs

klausagnoletti:
## What this does

Follow-up to #428. Adds tests for the merged SCSS plugin, fixes three runtime bugs found during review, and replaces the hallucinated documentation with an accurate README.

## Bug fixes

Three bugs were caught that would have caused the plugin to silently produce zero findings at runtime:

| Bug | File | Fix |
|-----|------|-----|
| `cmd` contained `{file_path}` — the runner passes it as a literal string, no substitution happens | `__init__.py` | Changed to glob patterns `"**/*.scss" "**/*.sass"`, matching every other generic plugin |
| `fmt: 'json'` doesn't match stylelint's output — stylelint emits `[{source, warnings}]`, but `parse_json` expects a flat `[{file, line, message}]` array, yielding zero parsed entries | `__init__.py` | Switched to `--formatter unix` + `fmt: 'gnu'`; `parse_gnu` handles `path:line:col: message` correctly |
| `timeout` and `validate_paths` are silently dropped by `normalize_tool_specs` — only `label/cmd/fmt/id/tier/fix_cmd` are retained | `__init__.py` | Removed the dead keys |

## Tests (8)

```
test_config_name
test_config_extensions
test_detect_markers
test_detect_commands_non_empty
test_has_stylelint_phase
test_exclusions
test_command_has_no_placeholder   ← reads captured cmd from closure, guards against {file_path} regression
test_parsing                      ← round-trips stylelint unix output through parse_gnu
```

## Documentation

Replaced the 455-line README (which referenced `SCSSConfig`, `desloppify report`, `--detailed`, and `install-tools` — none of which exist) with a 30-line version that only describes what the plugin actually does.

## How the bugs were found

This PR was produced with a multi-agent workflow: Claude wrote the initial cleanup and tests, Codex reviewed the result and identified all three runtime bugs, and Gemini independently verified the fixes, added the `test_parsing` round-trip test, and caught an incorrect `autofix` usage example in the README (`fixer` is a required positional argument).

Fairly mundane work, but the multi-agent part sounds fancy.

### DIFF
--- a/desloppify/languages/scss/README.md
+++ b/desloppify/languages/scss/README.md
@@ -1,0 +1,31 @@
+# SCSS Language Plugin for Desloppify
+
+Provides SCSS/Sass analysis via [stylelint](https://stylelint.io/).
+
+## Supported extensions
+
+`.scss`, `.sass`
+
+## Requirements
+
+```bash
+npm install -g stylelint stylelint-config-standard-scss
+```
+
+## Project detection
+
+The plugin activates when desloppify finds a `_scss` directory or a `.stylelintrc` file in the project.
+
+## Excluded paths
+
+`node_modules`, `_output`, `.quarto`, `vendor`
+
+## Usage
+
+```bash
+# Scan a project
+desloppify scan --path <project>
+
+# Auto-fix stylelint issues
+desloppify autofix stylelint-issue --path <project>
+```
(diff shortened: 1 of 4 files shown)


peteromallet:
Great follow-up — all three bug fixes are substantive (the `{file_path}` placeholder and wrong formatter would've meant zero findings at runtime). Tests look good. Waiting on #428 to land first so this has a clean base, then I'll merge. Thanks\!

PR #453, opened 2026-03-16
fix: Windows codex runner path quoting and log encoding

### LINKED ISSUE #442: `desloppify review --run-batches --runner codex` fails on Windows, then crashes during log recovery
## Summary

On Windows, `desloppify` fails to complete a Codex-backed subjective review batch, then crashes while trying to recover the failed batch log.

Observed on:
- `desloppify 0.9.9`
- Windows / PowerShell
- repo path containing spaces: `C:\Users\xxxxx\Documents\GitHub\yyyyyyy`

This appears to involve **two separate issues**:

1. The Codex batch runner fails to produce a valid JSON payload.
2. `desloppify` then crashes with a Windows text decoding error while reading the batch log.

## Versions

- Local `desloppify` before update: `0.9.8`
- Updated to latest published version on **March 15, 2026**: `0.9.9`
- Local Codex skill refreshed with:
  - `desloppify update-skill codex`

## Repro

Command:

```powershell
desloppify review --packet ".desloppify\review_packets\holistic_packet_20260315_173333.json" --run-batches --runner codex --only-batches 1

Actual Result
Console output:

Batch 1/1 started (#1)
Batch 1/1 exited (1); pending payload recovery in 9s (#1)
batch output JSON parse failed: no valid payload found
...
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 1437
Traceback ends in:

desloppify\app\commands\review\runner_process_impl\io.py", line 55, in extract_payload_from_log
log_text = log_path.read_text()
...
UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d

Expected Result
The Codex review batch should run and emit a valid JSON payload, or
if the subagent fails, desloppify should report the batch failure cleanly without crashing during log parsing.
Important Artifacts
Run directory from repro:

.desloppify\subagents\runs\20260315_184529
Relevant files:

run.log
logs\batch-1.log
What batch-1.log Shows
The spawned command is logged as:

cmd /c C:\Users\denys\AppData\Roaming\npm\codex.CMD exec --ephemeral -C C:\Users\denys\Documents\GitHub\ai_idle\core_project - Copy -s workspace-write -c approval_policy="never" -c model_reasoning_effort="low" -o C:\Users\denys\Documents\GitHub\ai_idle\core_project - Copy\.desloppify\subagents\runs\...\results\batch-1.raw.txt ...
Later in the same log:

WARNING: proceeding, even though we could not update PATH: Access is denied. (os error 5)
Error: Access is denied. (os error 5)
Suspected Problems
1. Windows quoting bug for paths with spaces
The logged cmd /c ... codex.CMD exec ... invocation does not quote the -C and -o paths.

Example problematic fragment:

-C C:\Users\denys\Documents\GitHub\ai_idle\core_project - Copy
Because the repo path contains spaces, this likely breaks argument parsing on Windows.

I also manually reproduced that an unquoted invocation fails differently with argument parsing issues like:

error: unexpected argument 'Copy' found
So at minimum, the runner should quote path arguments it passes to cmd /c.

2. Failure recovery reads logs using the wrong encoding on Windows
After the batch fails, desloppify attempts to recover/extract payload from the log and calls:

log_path.read_text()
On this Windows machine that resolves to cp1252, but the log contains bytes that require UTF-8, causing:

UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d ...
This turns an ordinary batch failure into a hard crash.

peteromallet:
## Summary
- Quote executable paths containing spaces when building `cmd /c` commands on Windows, preventing subprocess spawn failures
- Add `encoding="utf-8", errors="replace"` to `read_text()` calls in log recovery (`io.py`), preventing `UnicodeDecodeError` on Windows where the default encoding is cp1252
- Bump version to 0.9.10

Fixes #442

## Test plan
- [x] Full test suite passes (5364 passed, 3 skipped)
- [ ] Manual verification on Windows with a path containing spaces
- [ ] Manual verification with a log file containing non-cp1252 bytes

### DIFF
--- a/desloppify/app/commands/runner/codex_batch.py
+++ b/desloppify/app/commands/runner/codex_batch.py
@@ -4,6 +4,7 @@ from __future__ import annotations
 
 import os
 import shutil
+import subprocess
 import sys
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


peteromallet:
## Fix corrected: proper `cmd /c` argument quoting

The previous commit pre-quoted the **executable** path in `_resolve_executable`, but the actual breakage was in **argument** paths (`-C repo_root`, `-o output_file`) containing spaces. The error `unexpected argument 'Copy' found` confirms `-C core_project - Copy` was being split at the space.

### Why pre-quoting was wrong

`subprocess.Popen` with a list calls `list2cmdline()` which adds its own quoting. Pre-embedding `"` around the executable caused **double-quoting** — `list2cmdline` would produce `"\"C:\...\codex.CMD\""`.

### The real problem

`cmd /c` concatenates everything after `/c` into a single string and re-parses it with its own tokeniser. Even though `list2cmdline` quotes individual arguments with spaces, `cmd`'s re-parsing has edge cases where those quotes don't survive properly.

### The fix

Introduced `_wrap_cmd_c()` which:
1. Takes the full command list `["cmd", "/c", "codex.cmd", "exec", "-C", "path with spaces", ...]`
2. Uses `subprocess.list2cmdline()` on everything after `/c` to build one properly-quoted string
3. Returns `["cmd", "/c", "codex.cmd exec -C \"path with spaces\" ..."]`

This way, `cmd /c` receives exactly the command string we intended as a single token.

### What's kept
- The `encoding="utf-8", errors="replace"` fix in `io.py` is correct and unchanged.

### Tests
- Added `test_wrap_cmd_c_collapses_arguments_into_single_string` — verifies the collapsing logic
- Added `test_codex_batch_command_on_windows_collapses_cmd_c` — end-to-end test mocking Windows platform
- Full suite: 5366 passed, 3 skipped

PR #454, opened 2026-03-16
fix: allow scan when queue is fully drained (#441)

### LINKED ISSUE #441: scan refuses after next reports queue cleared
## Problem

In a repo with an active plan, `desloppify next --count 5` can report the queue is cleared and explicitly instruct `desloppify scan`, but `desloppify scan --path .` then immediately fails with the mid-cycle guard.

## Repro

Project: local monorepo scan on `desloppify` installed via pip/git.

Commands and outputs:

1. Run `desloppify next --count 5`

Output excerpt:

    Queue: 0 items

    Queue cleared!
    Frozen plan-start: strict 73.8 -> Live estimate: strict 73.8
    Run `desloppify scan` now to finalize and reveal your updated score.

2. Run `desloppify scan --path .`

Output excerpt:

    0 items remaining in your queue.
    Scanning mid-cycle regenerates issue IDs and breaks triage state.
    Work through items with `desloppify next`, then scan when clear.

## Expected

If `next` reports the queue is cleared and recommends `scan`, the normal `scan` command should succeed without requiring `--force-rescan`.

## Actual

`scan` still treats the repo as mid-cycle and refuses to run.

## Notes

This happened after resolving review items with `desloppify plan resolve ...` in a dirty working tree on macOS. The `next` / `scan` contradiction was reproducible immediately after the queue-cleared message.

peteromallet:
## Summary
- When the work queue drains to 0 items, the lifecycle phase may not advance to "scan", causing `score_display_mode()` to return `PHASE_TRANSITION` instead of `LIVE`
- The scan preflight guard then blocks the scan even though there's nothing left to do — a UX contradiction where `next` says "queue cleared" but `scan` refuses to run
- Added an early return in `scan_queue_preflight` when `breakdown.queue_total == 0`, bypassing the lifecycle phase check entirely

## Test plan
- [x] Added regression test `test_queue_drained_with_non_scan_lifecycle_allows_scan` covering the exact scenario (queue_total=0, lifecycle_phase="review")
- [x] All 5365 existing tests pass (3 skipped)

Fixes #441

### DIFF
--- a/desloppify/app/commands/helpers/queue_progress.py
+++ b/desloppify/app/commands/helpers/queue_progress.py
@@ -106,6 +106,8 @@ def score_display_mode(
         return ScoreDisplayMode.LIVE
     if breakdown is None:
         return ScoreDisplayMode.LIVE
+    if breakdown.queue_total == 0:
+        return ScoreDisplayMode.LIVE  # Queue fully drained — always live (#441)
     if breakdown.lifecycle_phase == LIFECYCLE_PHASE_SCAN:
         return ScoreDisplayMode.LIVE
     if breakdown.lifecycle_phase == LIFECYCLE_PHASE_EXECUTE:
--- a/desloppify/app/commands/scan/preflight.py
+++ b/desloppify/app/commands/scan/preflight.py
@@ -94,8 +94,6 @@ def scan_queue_preflight(args: object) -> None:
     except OSError:
         _logger.debug("scan preflight queue breakdown skipped", exc_info=True)
         return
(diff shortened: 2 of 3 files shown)


peteromallet:
Reworked — moved the `queue_total == 0` check into `score_display_mode()` itself instead of adding an ad-hoc bypass in preflight. This fixes the bug for all callers (status display, plan nudge, next flow), not just scan. Added parametrized tests covering review/execute/workflow lifecycle phases. Fixes #441.

Issue #446, opened 2026-03-16
SKILL.md assumes `pip install` — should be package-manager-agnostic


willfrey:
## Problem

The SKILL.md instructions assume `pip install desloppify[full]` as the installation method. In practice, projects use a variety of dependency management approaches:

- `uv add --dev desloppify[full]` (add as a dev dependency)
- `uv pip install desloppify[full]` (install into a uv-managed venv)
- `uvx desloppify[full]` (run without installing, which the CLI examples already use)
- `pipx install desloppify[full]`
- `poetry add --group dev desloppify[full]`
- etc.

Some users may not want desloppify added to their project dependencies at all and prefer ephemeral execution via `uvx`/`pipx run`. Others may want it in a specific dependency group. The current instructions don't account for this.

## Possible approaches

A few ideas, not mutually exclusive:

1. **Minimal fix — document alternatives**: Replace the `pip install` instruction with a short section showing common options, and let the user pick.

2. **Config-driven**: Add an optional field to `.desloppify/config.json` (e.g. `"run_prefix"` or `"command"`) that the skill instructions reference instead of hardcoding `pip install`. Projects using `uv` could set `"run_prefix": "uv run"`, and the SKILL.md could say "install/run desloppify using your project's preferred method."

3. **Auto-detect**: Check for `pyproject.toml` (uv/poetry), `Pipfile` (pipenv), etc. and adapt the instructions accordingly. Probably overkill.

4. **Just use `uvx` everywhere**: The CLI examples in SKILL.md already use `uvx desloppify[full] ...` for all commands, so the install step could simply be removed — `uvx` handles it. The only gap would be users who *want* it as a tracked dependency.

No strong opinion on which approach is best — just flagging that the current assumption trips up projects that don't use bare `pip`.

peteromallet:
Fair point re: pip install. Switched to uvx-first language in PR #455. Thanks!

peteromallet:
Fixed in 0.9.10 — switched to uvx-first install language in SKILL.md (PR #455). Thanks for the report!

Issue #445, opened 2026-03-16
SKILL.md batch file naming inconsistency causes agents to use `.json` instead of `.raw.txt`


willfrey:
## Problem

The SKILL.md file has conflicting signals about batch result file naming. The tool implementation (`runner_packets.py`, `orchestrator.py`) expects `batch-N.raw.txt`, and there's a `CRITICAL` note at line ~295 of SKILL.md that says so:

```
**CRITICAL**: Results files must be named `batch-N.raw.txt` (not `.json`).
The import command expects this exact naming convention.
```

However, multiple earlier and more prominent references in the same file use `.json`:

- Line ~123: `uvx desloppify[full] review --import merged.json`
- Line ~125: `uvx desloppify[full] review --import file.json`
- Line ~166: `uvx desloppify[full] review --import findings.json`
- Line ~167: `uvx desloppify[full] review --import findings.json`

## Impact

In practice, agents read the `.json` examples first and name their batch output files `batch-N.json`. The import command then fails because it's looking for `batch-N.raw.txt`. The agent has to spend extra turns diagnosing the error and renaming files, wasting tokens every time.

## Suggested Fix

Update the `.json` references in the review workflow examples to use `.raw.txt` (or whatever the canonical extension should be), so the SKILL.md is internally consistent and agents produce correctly-named files on the first try.

peteromallet:
Yeah the .json/.raw.txt inconsistency was definitely confusing agents. Cleaned up in PR #455, thanks for flagging.

peteromallet:
Fixed in 0.9.10 — batch file naming standardized to .raw.txt throughout SKILL.md (PR #455). Thanks for the report!

PR #455, opened 2026-03-16
fix(docs): SKILL.md cleanup — frontmatter, file naming, install

### LINKED ISSUE #444: `allowed-tools` frontmatter attribute not supported in Claude Code skill files
## Problem

The `SKILL.md` file uses `allowed-tools` in its frontmatter, but Claude Code does not support this attribute in skill files. This produces a warning:

> Attribute 'allowed-tools' is not supported in skill files. Supported: argument-hint, compatibility, description, disable-model-invocation, license, metadata, name, user-invocable.

## Suggested Fix

Remove `allowed-tools` from the skill frontmatter, or replace it with a supported attribute if the intent is to restrict which tools the skill can use.

peteromallet:
## Summary
- **#444**: Remove unsupported `allowed-tools: Bash(desloppify *)` frontmatter from SKILL.md — Claude Code doesn't support this attribute
- **#445**: Specify `batch-N.raw.txt` naming in overlay docs (CLAUDE.md, HERMES.md, CURSOR.md, WINDSURF.md) to match what the batch orchestrator actually expects
- **#446**: Replace `pip install` with `uvx` in SKILL.md prerequisite check, consistent with existing CLI examples

Also bumps version to 0.9.10.

## Test plan
- [ ] Verify SKILL.md frontmatter parses without `allowed-tools` line
- [ ] Confirm `uvx --from git+https://github.com/peteromallet/desloppify.git desloppify` installs correctly
- [ ] Verify overlay docs now reference `.raw.txt` consistently

Fixes #444, #445, #446

### DIFF
--- a/.agents/skills/desloppify/SKILL.md
+++ b/.agents/skills/desloppify/SKILL.md
@@ -6,11 +6,10 @@ description: >
   duplicate functions, code smells, naming issues, import cycles, or coupling
   problems. Also use when asked for a health score, what to fix next, or to
   create a cleanup plan. Supports 29 languages.
-allowed-tools: Bash(desloppify *)
 ---
 
 <!-- desloppify-begin -->
-<!-- desloppify-skill-version: 5 -->
+<!-- desloppify-skill-version: 6 -->
 
 # Desloppify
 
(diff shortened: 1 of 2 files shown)


willfrey:
Just a FYI, you may want to update the template you provide in the README.md since it still references `pip install`, etc.

Thanks for addressing this so quickly!

peteromallet:
Reworked — the original version missed the `.agents/skills/desloppify/SKILL.md` copy (git-tracked, used by Codex). Now both copies are updated and synced: `allowed-tools` removed, uvx-first with pip fallback, explicit batch naming note (`batch-N.raw.txt` not `.json`), version bumped to 6. Thanks @willfrey for flagging the README gap — that's covered now too.

willfrey:
One other nitpick. `uvx` won't install it to your path. You have to execute it via `uvx 'desloppify[full]'` each time. One alternative is `uv tool install 'desloppify[full]'` to have it in your path as an executable.

Edit: noticed the check in the SKILL.md file that just looks for `desloppify` in the `PATH`.

PR #456, opened 2026-03-16
fix: merge retry batch results with original run before coverage check

### LINKED ISSUE #443: review --run-batches retry for failed slice does not merge with original run coverage
## Summary
When a `desloppify review --run-batches --runner codex --parallel --scan-after-import` run completes with one failed batch, retrying only the failed batch via the immutable packet path succeeds, but the retry cannot be imported because coverage is evaluated against only the retried slice instead of the original run plus retry.

## Repro
Repo under review: `/Users/andrew/provenance`

1. Run:
```bash
desloppify review --run-batches --runner codex --parallel --scan-after-import
```
2. Let the run finish with one failed batch. In my case the command ended with:
```text
Failed batches: [7]
Retry command:
  desloppify review --run-batches --packet /Users/andrew/provenance/.desloppify/review_packets/holistic_packet_20260316_125314.json --only-batches 7
```
3. Retry only that batch:
```bash
desloppify review --run-batches --packet /Users/andrew/provenance/.desloppify/review_packets/holistic_packet_20260316_125314.json --only-batches 7
```

## Expected
The retry should merge with the original run artifacts and allow a trusted import/scan for the full 20-dimension review set.

## Actual
The retry finishes the missing batch, but import fails closed with incomplete coverage for the full review set:
```text
Coverage gap: imported assessments for 1/20 scored subjective dimensions.
Still missing: naming_quality, logic_clarity, type_safety, contract_coherence, error_consistency, +14 more
Retry with full coverage or explicitly bypass with --allow-partial.
Error: trusted assessment import blocked due to incomplete selected-dimension coverage.
```

This makes the documented packet retry path unusable for failed-slice recovery, because the successful retried slice is not reconciled with the already completed batches from the original run.

## Notes
This was on a repo where the objective queue was already clear, so the failure mainly blocks honest subjective score refresh after a large fix batch.

peteromallet:
## Summary
- When retrying failed batches with `--only-batches N --packet ...`, the import now searches for the original run's merged results and combines them with the retry's results before evaluating the coverage gate.
- Prior to this fix, retried batches were checked against the full dimension set in isolation (e.g. 1/20 covered), causing the import to be rejected even though the original run had successfully covered the other 19 dimensions.
- Added `find_prior_run_merged_results` to locate the most recent prior run using the same immutable packet, and `overlay_retry_results_on_prior` to merge assessments/issues/notes with retry results taking precedence for retried dimensions.

Fixes #443

## Test plan
- [ ] Run existing test suite (`python -m pytest desloppify/tests/ -q -x`) -- all 5364 tests pass
- [ ] Manual test: run a full batch review, then retry a single failed batch with `--only-batches` and verify the import succeeds with merged coverage
- [ ] Verify that a full run (no `--only-batches`) is unaffected by the change

### DIFF
--- a/desloppify/app/commands/review/batch/execution_phases.py
+++ b/desloppify/app/commands/review/batch/execution_phases.py
@@ -559,19 +559,6 @@ def _is_partial_batch_retry(prepared: PreparedBatchRunContext) -> bool:
     return set(prepared.selected_indexes) != all_indexes
 
 
-def _retry_dimension_set(prepared: PreparedBatchRunContext) -> set[str]:
-    """Return the set of dimensions covered by the selected (retried) batches."""
-    dims: set[str] = set()
-    for idx in prepared.selected_indexes:
-        if 0 <= idx < len(prepared.batches):
-            batch = prepared.batches[idx]
-            if isinstance(batch, dict):
-                for dim in batch.get("dimensions", []):
-                    if isinstance(dim, str) and dim.strip():
-                        dims.add(dim.strip())
-    return dims
-
-
 def merge_and_import_batch_run(
     *,
     prepared: PreparedBatchRunContext,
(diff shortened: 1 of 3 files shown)


peteromallet:
Reworked completely. The original merge-based approach (195 lines) had multiple edge cases: wrong prior-run selection, dimension normalization mismatches, stale metadata, zero tests. Replaced with a ~5-line fix — when it's a partial batch retry, just set `allow_partial = True` before the coverage gate. Same effect, no filesystem scanning, no result merging, no edge cases. 4 new tests. Fixes #443.

PR #449, opened 2026-03-16
feat(r): improve tree-sitter R_SPEC function and import queries

sims1253:
## Summary

 Enhances the shared tree-sitter R_SPEC to capture additional R patterns:

 - **Anonymous function detection**: Captures function definitions passed as arguments to calls
 like `lapply()`, `purrr::map()`, `sapply()`, etc.
 - **Namespace operator capture**: Recognizes `pkg::fn` patterns (e.g., `dplyr::select`,
 `data.table::fread`) as imports

 Both patterns capture `@path` for compatibility with the dependency graph and unused imports
 analysis.

 ## Files changed

 - `desloppify/languages/_framework/treesitter/specs/scripting.py` (+12 lines)

 ## Testing

 Existing tree-sitter tests continue to pass. The new patterns integrate seamlessly with the
 existing R_SPEC.

 ## Dependencies

 None - this is a standalone improvement.

PR #461, opened 2026-03-16
fix: extract anonymous functions in tree-sitter specs

peteromallet:
## Summary
- **Bug**: PR #449 added an R tree-sitter query pattern for anonymous functions (e.g., `lapply(x, function(i) {...})`) that captures `@fn` but the function extractor requires `@name`. All anonymous function matches were silently skipped.
- **Fix**: Updated the extractor to synthesize an `<anonymous>` name when `@name` is absent but `@func` is present, rather than skipping the match entirely.
- **Tests**: Added 3 tests for R function extraction covering named functions, anonymous function detection, and metadata correctness.

Credits to @sims1253 for the original R spec in PR #449.

## Test plan
- [x] New `TestRExtraction` tests pass (named + anonymous function extraction)
- [x] Full test suite passes (5367 passed, 3 skipped)
- [ ] Manual: scan an R project with `lapply`/`sapply` anonymous functions and verify they appear in results

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "desloppify"
-version = "0.9.9"
+version = "0.9.10"
 description = "Multi-language codebase health scanner and technical debt tracker"
 readme = "README.md"
 requires-python = ">=3.11"


peteromallet:
Follow-up to PR #449 by @sims1253. The anonymous function tree-sitter pattern captured `@fn` but the extractor requires `@name`, so all matches were silently skipped. Fixed the extractor to fall back to `<anonymous>` when `@name` is absent. Added 3 tests for R function extraction. Credit to sims1253 for the original query pattern.

PR #472, opened 2026-03-17
fix: handle dataclass serialization in json_default

ammonwk:
## Problem
`desloppify scan` crashes during state persistence when `EcosystemFrameworkDetection` (a dataclass) is stored in scan state. `json_default` in `schema_scores.py` doesn't handle dataclass instances, causing:

```
TypeError: Object of type EcosystemFrameworkDetection is not JSON serializable: EcosystemFrameworkDetection(ecosystem='node', ...)
```

## Fix
Added a `dataclasses.is_dataclass()` check to `json_default` that converts dataclass instances to dicts via `dataclasses.asdict()`, matching the existing pattern for other non-JSON-native types (sets, Paths, datetime objects).

### DIFF
--- a/desloppify/engine/_state/schema_scores.py
+++ b/desloppify/engine/_state/schema_scores.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


peteromallet:
Thanks @ammonwk! You correctly identified the dataclass serialization crash in `json_default` — that's a real bug. We implemented the same fix independently (commit 61bb7cb3 on `0.9.11`) as part of triaging several PRs that all found this issue around the same time.

The root cause has also been fixed separately (commit 986f8842, from PR #483) by keeping framework caches out of persisted state entirely.

Appreciate you finding this and submitting a fix!

peteromallet:
Thanks @ammonwk for the contribution/report. Final decision: reject. This dataclass serialization crash has already been fixed independently; this PR is now obsolete.

PR #475, opened 2026-03-17
Fix scan crash when framework cache hits persisted review_cache

MacHatter1:
### Problem
`desloppify scan` could crash while persisting state if `state["review_cache"]` contains framework detection / info dataclass instances (non-JSON-serializable).

### Fix
- Store JSON-safe dict payloads in `review_cache` for:
  - `frameworks.ecosystem.present:*` (framework detection)
  - `framework.nextjs.info:*` (Next.js info)
  - Decode these payloads back to the expected runtime dataclasses when read.
- Harden `json_default` to encode dataclasses/enums defensively so future cache regressions don't brick scans.

### Tests
- `pytest -q` (6236 passed, 165 skipped)
- Manual scans using workspace code via `PYTHONPATH`:
  - `vercel/nextjs-postgres-auth-starter`
  - `vercel/swr`
  - `sindresorhus/ky`

### DIFF
--- a/desloppify/engine/_state/schema_scores.py
+++ b/desloppify/engine/_state/schema_scores.py
@@ -2,12 +2,19 @@
 
 from __future__ import annotations
 
+from dataclasses import asdict, is_dataclass
+from enum import Enum
 from pathlib import Path
 from typing import Any
 
 
 def json_default(obj: Any) -> Any:
     """JSON serializer fallback for known non-JSON-native state values."""
+    if is_dataclass(obj) and not isinstance(obj, type):
+        return asdict(obj)
+    if isinstance(obj, Enum):
+        value = obj.value
+        return value if isinstance(value, str | int | float | bool | None) else obj.name
     if isinstance(obj, set):
         return sorted(obj)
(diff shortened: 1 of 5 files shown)


peteromallet:
Thanks @MacHatter1! You correctly identified the framework cache serialization crash — that's a real bug. The `json_default` dataclass handler part of your PR was implemented independently (commit 61bb7cb3 on `0.9.11`), and the root cause was separately fixed by PR #483 (commit 986f8842 by @maciej-trebacz) which introduces a `runtime_cache` to keep framework objects out of persisted state entirely.

Your encode/decode helpers for framework cache round-tripping addressed a secondary issue (cache miss after reload), but since PR #483's approach means framework objects never reach the persistence layer, that code path is no longer needed.

Appreciate the thorough investigation — you clearly dug deep into the serialization chain!

peteromallet:
Thanks @MacHatter1 for the contribution/report. Final decision: reject. This tolerates framework cache leakage after persistence, but the correct fix is to keep runtime framework caches out of persisted review state.

PR #477, opened 2026-03-17
fix: prevent catastrophic backtracking in dart function extractor regex

AvoMandjian:
## Problem
The `desloppify scan` command gets stuck and hangs indefinitely at the `[3/8] Signature analysis...` step when analyzing certain Dart files.

This is caused by catastrophic backtracking in the `_FUNC_DECL_RE` regex within `desloppify/languages/dart/extractors.py`. The nested quantifiers `(?:(?:@[\w\.\(\),<>\s]+)\s*)*` can cause exponential backtracking if a string contains multiple matches for the inner group (such as SQL strings containing `@` parameters) separated by whitespace, because both `[\w\.\(\),<>\s]+` and `\s*` can consume whitespace characters, leading to massive evaluation times.

## Fix
Replaced the non-possessive quantifiers `+` and `*` with possessive quantifiers `++` and `*+` in `_FUNC_DECL_RE` for all parts of the regex that consume characters safely. This guarantees that once a string is matched by the inner pattern, the regex engine won't backtrack and try splitting up the whitespace, thereby instantly failing or matching the text without any performance penalties.

Verified that tests pass and the dart project now completes its scan properly.

### DIFF
--- a/desloppify/languages/dart/extractors.py
+++ b/desloppify/languages/dart/extractors.py
@@ -19,10 +19,10 @@ DART_FILE_EXCLUSIONS = ["build", ".dart_tool", ".fvm", ".git", "node_modules"]
 
 _FUNC_DECL_RE = re.compile(
     r"(?m)^\s*"
-    r"(?:(?:@[\w\.\(\),<>\s]+)\s*)*"
-    r"(?:(?:external|static|final|const|factory|late|required|covariant)\s+)*"
-    r"(?:[A-Za-z_]\w*(?:<[^>{}]+>)?\??\s+)?"
-    r"([A-Za-z_]\w*)\s*\(([^)]*)\)\s*(?:async\s*)?(=>|\{)"
+    r"(?:(?:@[\w\.\(\),<>\s]++)\s*+)*"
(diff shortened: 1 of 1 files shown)


peteromallet:
Thanks @AvoMandjian! Great catch on the catastrophic backtracking — the overlapping `\s` in the annotation character class and trailing `\s*` created exponential blowup that could hang indefinitely on annotated Dart functions.

Cherry-picked into the `0.9.11` release branch (commit ab36319a). The possessive quantifier fix is clean and also fixes a latent correctness bug where annotated functions like `@override Widget build(...)` produced garbage names.

Really appreciate you tracing this down and testing it! 🎉

peteromallet:

peteromallet:
Thanks @AvoMandjian. I accepted this change.

Decision: implemented already. The review confirmed the catastrophic-backtracking problem in the Dart function extractor regex was real, and the possessive-quantifier fix is the right narrow fix for Python 3.11+. The accepted change is already present in this integration branch as commit ab36319a, so I did not need to make a new code change.

I also ran a focused sanity check confirming annotated Dart methods still capture the real function name and @-heavy non-matching input fails cleanly.

Leaving the PR closed because the fix is already carried in the branch.

PR #478, opened 2026-03-17
fix: clear stale test coverage issues on zone reclassification

claytona500:
## Problem
`desloppify` was surfacing stale `test_coverage` findings for files that had already been reclassified into non-production zones, including `.test.js` files in the current project. The detector logic itself was not necessarily re-emitting those findings; they were persisting through state merge/reconciliation.

## Fix
- pass live `zone_lookup` into scan merge reconciliation
- auto-resolve open `test_coverage` issues when the current file zone is non-production (`test`, `config`, `generated`, `vendor`)
- harden JavaScript zone rules so `.test.`, `.spec.`, and `__tests__` classify as `test`
- add focused tests for reconciliation and JS zone classification

## Verification
- `python -m py_compile` on the touched Python files succeeded under the local 3.12 env
- custom reconciliation check confirmed stale open `test_coverage` issues resolve when `zone_lookup` returns `test`
- simulated merge against a real project state reduced open `.test.js` `test_coverage` issues from 108 to 0

### DIFF
--- a/desloppify/app/commands/scan/workflow.py
+++ b/desloppify/app/commands/scan/workflow.py
@@ -443,6 +443,7 @@ def merge_scan_results(
 
     if runtime.lang and runtime.lang.zone_map is not None:
         runtime.state["zone_distribution"] = runtime.lang.zone_map.counts()
+    zone_lookup = runtime.lang.zone_map.get if runtime.lang and runtime.lang.zone_map is not None else None
     _persist_scan_coverage(runtime.state, runtime.lang)
 
     target_score = target_strict_score_from_config(runtime.config)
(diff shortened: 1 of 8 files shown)


peteromallet:
Thanks @claytona500! You identified a real problem — stale test_coverage findings persisting after zone reclassification is a genuine gap in verify_disappeared. The JS zone rules you wrote (`.test.`, `.spec.`, `__tests__/`, `__mocks__/` patterns) are also solid.

We implemented a slightly different approach (commit 34e7f1d8 on `0.9.11`): instead of hardcoding `test_coverage` specifically, we use the existing `should_skip_issue()` from `zones.py` which checks `ZONE_POLICIES` generically. This means any detector/zone combination where policy says "skip" will auto-resolve stale findings — not just test_coverage. Your JS zone rules were taken directly and credited in the code.

Appreciate the thorough investigation — both the bug report and the zone rules made this fix possible! 🎉

peteromallet:

PR #481, opened 2026-03-18
fix: handle dataclass objects in json_default serializer

ucanlendit:
## Problem
`desloppify scan` crashes during `save_state` when `EcosystemFrameworkDetection` (a dataclass) is in the state dict. The `json_default` fallback in `schema_scores.py` doesn't handle dataclass instances.

```
TypeError: Object of type EcosystemFrameworkDetection is not JSON serializable: EcosystemFrameworkDetection(ecosystem='node', package_root=PosixPath('...'), package_json_relpath='package.json', present={})
```

Command: `desloppify scan --path .` on a Node/TypeScript monorepo.

## Fix
Added `dataclasses.is_dataclass()` check to `json_default()` that converts dataclass instances to dicts via `dataclasses.asdict()`, before the `TypeError` raise.

### DIFF
--- a/desloppify/engine/_state/schema_scores.py
+++ b/desloppify/engine/_state/schema_scores.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+import dataclasses
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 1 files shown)


peteromallet:
Thanks for the contribution @ucanlendit! You correctly identified the dataclass serialization crash — that's a real bug. We ended up implementing the same fix independently (commit 61bb7cb3 on `0.9.11` — a `dataclasses.asdict()` handler in `json_default`) as part of triaging several related PRs that all found the same issue.

The root cause has also been addressed in a separate commit (986f8842, from PR #483 by @maciej-trebacz) which separates runtime caches from persisted state entirely.

Appreciate you taking the time to investigate and submit a fix! 🎉

peteromallet:
Thanks @ucanlendit for the contribution/report. Final decision: reject. This duplicates the dataclass serialization fix that has already shipped independently, so we will not merge another version.

PR #484, opened 2026-03-18
fix: handle generic fixers that return entries without 'removed' key

AugusteBalas:
## Problem

Running `desloppify autofix eslint-warning` (and any other generic fixer using `fix_cmd`) crashes with:

```
  File ".../desloppify/app/commands/autofix/cmd.py", line 40, in cmd_autofix
    total_items = sum(len(r["removed"]) for r in results)
KeyError: 'removed'
```

**Root cause**: `make_generic_fixer` in `tool_factories.py` returns `FixResult` entries with shape `{"file": ..., "line": ...}` (dry-run) or `{"file": ..., "fixed": True}` (actual fix), whereas TypeScript-style fixers return `{"removed": [...symbols], "lines_removed": N}`. Four places in the autofix pipeline assumed `"removed"` is always present.

## Fix

Guard all four access sites with `.get("removed", [])` / `"removed" in r` checks:

- `cmd.py`: count each entry as 1 item when `"removed"` is absent
- `apply_flow.py`: fall back to `"fixed"` label in the per-file summary
- `preview.py`: guard the count sum and `set()` construction
- `apply_retro.py`: guard symbol iteration and count/lines aggregation

## Verification

```
python3 -m pytest desloppify/tests/ -q
# 5438 passed, 3 skipped

python3 -m desloppify autofix eslint-warning --path <project> --dry-run
# Works correctly now
```

Tested against a real project with 154 eslint_warning findings.

### DIFF
--- a/desloppify/app/commands/autofix/apply_flow.py
+++ b/desloppify/app/commands/autofix/apply_flow.py
@@ -59,9 +59,13 @@ def _print_fix_summary(
         )
     )
     for result in results[:30]:
-        symbols = ", ".join(result["removed"][:5])
-        if len(result["removed"]) > 5:
-            symbols += f" (+{len(result['removed']) - 5})"
+        removed = result.get("removed", [])
+        if removed:
+            symbols = ", ".join(removed[:5])
+            if len(removed) > 5:
+                symbols += f" (+{len(removed) - 5})"
+        else:
+            symbols = result.get("summary", "fixed")
         extra = f"  ({result['lines_removed']} lines)" if result.get("lines_removed") else ""
         print(f"  {rel(result['file'])}{extra}  →  {symbols}")
     if len(results) > 30:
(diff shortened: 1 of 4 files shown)


peteromallet:
Thanks @AugusteBalas! Great catch — the `KeyError` on generic fixers was a real bug that would crash `desloppify autofix` for any non-TypeScript-style fixer.

Cherry-picked your fix into the `0.9.11` release branch (commit 18b497a2) with two additions: regression tests for the generic fixer result shape, and a code comment crediting you. The devil's advocate review also noted that the downstream pipeline doesn't fully handle generic fixer results (dry-run shows no samples, `_resolve_fixer_results` silently skips them) — we'll file a follow-up to normalize the schema at the source in `make_generic_fixer`.

Appreciate the contribution! 🎉

peteromallet:

peteromallet:

PR #485, opened 2026-03-18
fix: handle framework dataclass cache + synthetic deferred loop

ryexLLC:
## Problem
- `desloppify scan` can crash while saving state when framework detection objects are present in `review_cache` with:
  - `TypeError: Object of type EcosystemFrameworkDetection is not JSON serializable`
- `workflow::deferred-disposition` can reappear indefinitely when temporary-skipped entries include synthetic IDs (workflow/triage), creating a phantom deferred loop.

## Fix
- Extend JSON fallback serialization in `schema_scores.json_default()` to handle dataclass instances via `dataclasses.asdict(...)`.
- Filter synthetic IDs out of deferred temporary-skip counting in `build_deferred_disposition_item()` so only real issue IDs participate.
- Add regression tests for both behaviors:
  - dataclass serialization in state score schema helpers
  - synthetic-skip exclusion in deferred workflow item generation

## Verification
- `uv run --directory /tmp/desloppify-fix --with pytest pytest /tmp/desloppify-fix/desloppify/tests/engine/test_state_schema_scores_direct.py /tmp/desloppify-fix/desloppify/tests/review/test_work_queue_synthetic_workflow_direct.py -q`
- `uv run --directory /tmp/desloppify-fix desloppify scan --path /Users/alexanderting/Documents/frquick`
  - completes without the prior serialization crash
- `uv run --directory /tmp/desloppify-fix desloppify next --format json --count 1`
  - no phantom `workflow::deferred-disposition` loop in this run

### DIFF
--- a/desloppify/engine/_state/schema_scores.py
+++ b/desloppify/engine/_state/schema_scores.py
@@ -2,6 +2,7 @@
 
 from __future__ import annotations
 
+from dataclasses import asdict, is_dataclass
 from pathlib import Path
 from typing import Any
 
@@ -12,6 +13,8 @@ def json_default(obj: Any) -> Any:
         return sorted(obj)
     if isinstance(obj, Path):
         return str(obj).replace("\\", "/")
+    if is_dataclass(obj):
+        return asdict(obj)
     if hasattr(obj, "isoformat"):
         return obj.isoformat()
     raise TypeError(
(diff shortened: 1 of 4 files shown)


peteromallet:
Thanks @ryexLLC! Your PR had two fixes and we cherry-picked the synthetic deferred loop fix (commit 1a0439e2 on `0.9.11`) — great work tracing the code path through `migrate_deferred_to_skipped()` and `skip_items()`. The `is_synthetic_id()` filter is exactly the right approach, matching the pattern used in 6+ other locations.

For the dataclass serialization part: we handled that separately with a simpler `dataclasses.asdict()` approach (commit 61bb7cb3), prompted by the same bug report from PR #486. Your contribution to both fixes is credited in the code.

Appreciate the thorough investigation and the regression tests! 🎉

peteromallet:

peteromallet:

PR #495, opened 2026-03-20
fix: force UTF-8 in review runner logs and payload parsing

pietrondo:
## Problem
On Windows/Codex review runs, the batch runner and failure reporter still use the platform default text encoding in a few places. That causes two concrete failures:

- live stream/log reads can throw `charmap` decode errors while Codex emits UTF-8
- payload and log recovery can misread UTF-8 batch output, which then surfaces as malformed JSON or crashes in failure reporting

In a real `desloppify review --run-batches --runner codex --parallel --scan-after-import` run against `zoteropit`, this showed up as:
- `batch output JSON parse failed: Expecting ',' delimiter`
- `UnicodeDecodeError: 'charmap' codec can't decode byte ...` in `runner_failures.py`

After forcing UTF-8 decoding locally, the same retry path no longer crashed in parsing/reporting and correctly surfaced the real environment issue (`usage limit`) instead of a tool failure.

## Fix
- force `encoding="utf-8", errors="replace"` in the subprocess runner paths
- read runner logs with UTF-8 replacement in failure classification/hint generation
- read raw batch payload files with UTF-8 replacement during result collection

## Verification
- `python -m pytest desloppify/tests/review/test_runner_internals.py -q`
- `python -m pytest desloppify/tests/review/test_runner_parallel_internals.py -q`
- `python -m pytest desloppify/tests/review/review_commands_runner_cases.py -q`

All passed locally.

### DIFF
--- a/desloppify/app/commands/review/runner_failures.py
+++ b/desloppify/app/commands/review/runner_failures.py
@@ -144,7 +144,9 @@ def summarize_failure_categories(*, failures: list[int], logs_dir: Path) -> dict
             category = "missing_log"
         else:
             try:
-                category = classify_runner_failure(log_file.read_text())
+                category = classify_runner_failure(
+                    log_file.read_text(encoding="utf-8", errors="replace")
+                )
             except OSError:
                 category = "log_read_error"
         categories[category] = categories.get(category, 0) + 1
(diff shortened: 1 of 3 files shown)


peteromallet:
Thanks for the fix, @pietrondo! The UTF-8 encoding issue on Windows is real and your approach (explicit `encoding="utf-8"` + `errors="replace"`) is exactly right for log/diagnostic reads.

Cherry-picked the file-read changes (runner_failures.py, runner_parallel/__init__.py) into `0.9.12` as commit `8411efed`. Dropped the subprocess encoding changes in attempts.py — those go through a test-injection seam and the runner process already emits UTF-8, so the file reads are where the value is.

Thanks again for the clean, surgical fix!

peteromallet:

PR #497, opened 2026-03-21
docs: optimize agent install instructions for copy-paste into agents

andrewwhitecdw:
Enhanced agentic install prompt with feedback from Perplexity.AI and GLM-5.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,16 +19,34 @@ Currently supports 29 languages — full plugin depth for TypeScript, Python, C#
 Paste this prompt into your agent:
 
 ```
-I want you to improve the quality of this codebase. To do this, install and run desloppify.
-Run ALL of the following (requires Python 3.11+):
+You are a code quality agent using desloppify. Your mission: maximize the strict score
+through disciplined iteration. Never cut corners — complete large refactors and small
+fixes with equal rigor. Follow agent instructions precisely; do not substitute your
+own analysis. Maintain open-source toolchain discipline.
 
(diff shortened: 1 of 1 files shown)


peteromallet:
Thanks for the contribution, @andrewwhitecdw.

I'm going to pass on this one. The existing agent install prompt is deliberately tuned — the tone, the specific directives (like "don't be lazy" and THE LOOP explanation), and the backlog guidance are all intentional. The PR removes that content and replaces it with more generic AI-instruction patterns (e.g., "You are a code quality agent", "Maintain open-source toolchain discipline") that don't match how the tool actually works.

A few specific concerns:
- The hardcoded `desloppify exclude vendor build dist node_modules .git` assumes a directory structure that won't apply to most projects
- Directives like "Show before/after diffs" and "Confirm resolution command when complete" aren't part of desloppify's workflow
- The CORE LOOP section duplicates the scan/next/fix cycle that already appears in the prompt

I appreciate the effort, but this is one of those areas where the current version is intentionally non-standard. Thanks for thinking about the agent experience though.

peteromallet:
One more thing worth mentioning — a lot of the detailed agent workflow guidance (THE LOOP, backlog usage, resolve commands, etc.) is delivered via `desloppify update-skill`, which installs a full skill file tailored to each agent (Claude, Cursor, Copilot, Codex, etc.). So the README prompt is intentionally kept lightweight as an entry point — the heavy lifting happens in the skill file once installed.

PR #498, opened 2026-03-22
fix: test_coverage detector — naming mapping + importer count

Imwm82:
## Summary

Fixes two bugs in the test_coverage detector:

### Bug 1: map_test_to_source (TypeScript)

The first loop compared candidate basenames against the **TEST** file basename instead of the **SOURCE** basename, so the comparison always failed and the function skipped to the fallback loop.

```python
# Before (broken): compares candidate against test basename → never matches
for prod in production_set:
    prod_base = os.path.basename(prod)
    for c in candidates:
        if os.path.basename(c) == prod_base:  # always false
            return prod

# After (fixed): compare candidate basename against source basename
for c in candidates:
    c_base = os.path.basename(c)
    for prod in production_set:
        if c_base == os.path.basename(prod):  # correct
            return prod
```

### Bug 2: importer_count

Only counted production-file importers, ignoring test-file importers entirely. Combined both for accurate blast-radius assessment.

## Impact

For the CoQ Tracking System project:
- Test health: 3.7% -> 73.8%
- Strict score: 81.7 -> 86.8
- Test issues: 40 open -> 11 open (all genuine)

### DIFF
--- a/desloppify/engine/detectors/test_coverage/_issue_generation.py
+++ b/desloppify/engine/detectors/test_coverage/_issue_generation.py
@@ -12,6 +12,26 @@ from ._issue_quality import select_direct_test_quality_issue
 from .metrics import _file_loc, _loc_weight
 
 
+def _build_test_importer_index(
+    parsed_imports_by_test: dict[str, set[str]],
+) -> dict[str, set[str]]:
+    """Build a reverse index: production_file -> set of test files that import it."""
+    index: dict[str, set[str]] = {}
+    for test_path, prod_files in parsed_imports_by_test.items():
+        for prod_file in prod_files:
+            index.setdefault(prod_file, set()).add(test_path)
+    return index
+
+
+def _combined_importer_count(
+    filepath: str, graph: dict, test_importer_index: dict[str, set[str]]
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks for looking into the test_coverage detector, @Imwm82.

After reviewing the diff carefully, I'm not going to merge this — here's why:

**Bug 1 (map_test_to_source):** The existing code at `test_coverage.py:233-237` iterates `production_set`, sets `prod_base = os.path.basename(prod)`, then checks `os.path.basename(c) == prod_base`. This already correctly compares candidate basenames against production basenames. The PR's change swaps the loop nesting order and removes the redundant `prod in production_set` check, but the end result is semantically equivalent — it's a no-op refactor rather than a bug fix.

**Bug 2 (importer_count):** The existing code intentionally counts only production-file importers for blast-radius assessment. Adding test-file importers would inflate priority scores for well-tested files — the opposite of what we want. A file that's heavily tested should not appear riskier because of that.

If you're seeing unexpected test_health scores in your project, I'd be happy to look at a concrete example — there may be a different root cause. Thanks again for the investigation.

PR #540, opened 2026-04-10
fix: align queue builders with resolved subjective threshold

iatzmon:
## Problem
`desloppify next` and `desloppify plan queue` could disagree about the live queue when callers relied on the default subjective threshold.

In my repro, `next` passed a resolved target strict score through a `QueueContext`, while `plan queue` relied on the default `QueueBuildOptions.subjective_threshold = 100`. Because `build_execution_queue()` and `build_backlog_queue()` did not propagate the resolved threshold into the queue options, the snapshot builder used different thresholds for subjective visibility and phase selection.

That produced inconsistent queue surfaces for the same plan/state.

## Fix
- add `_resolved_subjective_threshold()` in `engine/planning/queue_policy.py`
- when queue builders synthesize or receive a `QueueContext`, copy the resolved target into `options.subjective_threshold` when the caller did not override it explicitly
- cover the regression with direct tests for both execution and backlog queue builders

## Verification
- `python -m pytest desloppify/tests/plan/test_plan_modules_direct.py -q`
- `python -m pytest desloppify/tests/engine/test_queue_context.py desloppify/tests/review/test_work_queue_plan_order_and_triage.py desloppify/tests/commands/test_next_queue_flow_direct.py -q`
- verified against a real project repro where `next` and `plan queue` previously diverged under the same JS state

### DIFF
--- a/desloppify/engine/planning/queue_policy.py
+++ b/desloppify/engine/planning/queue_policy.py
@@ -34,6 +34,23 @@ def _queue_plan_from_options(options: QueueBuildOptions) -> dict | None:
     return options.plan
 
 
+def _resolved_subjective_threshold(
+    state: StateModel,
+    options: QueueBuildOptions,
+) -> float:
+    """Return the effective subjective threshold for queue builders.
+
+    When callers pass a precomputed QueueContext, the queue builder must honor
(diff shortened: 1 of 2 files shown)


iatzmon:
Closing this PR per request.

PR #571, opened 2026-04-17
Fix plan migration dropping uncommitted findings

tristanmanchester:
## Summary
- make `_rename_key()` perform a true overwrite rename so legacy values are not discarded when the target key already exists
- run plan schema upgrades before applying empty-plan defaults so migrations match the state schema flow
- add regression tests for both the helper overwrite case and the runtime `resolve_plan_load_status()` path

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/plan/test_schema_migrations.py desloppify/tests/plan/test_persistence_runtime_paths.py`

### DIFF
--- a/desloppify/engine/_plan/schema/__init__.py
+++ b/desloppify/engine/_plan/schema/__init__.py
@@ -289,10 +289,10 @@ def ensure_plan_defaults(plan: dict[str, Any]) -> None:
 
     Runtime contract is v8-only. Legacy payloads are upgraded in-place once.
     """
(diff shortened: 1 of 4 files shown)


peteromallet:

I applied the plan migration change as reviewed. Focused plan migration tests passed, and the full suite passed. Local commit: `fbb4f36f`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as fbb4f36f.

PR #572, opened 2026-04-17
Fix tree-sitter dependency graph path mismatch

tristanmanchester:
## Summary
- normalize tree-sitter dep-graph `file_list` entries to absolute paths before parsing, graph initialization, and in-repo edge checks
- align graph keys with the absolute-path contract used by other dependency graph builders
- add regression coverage for production-style relative file lists returned by `find_source_files`

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/lang/common/test_treesitter_imports_direct.py desloppify/tests/lang/common/test_treesitter.py`

### DIFF
--- a/desloppify/languages/_framework/treesitter/imports/graph.py
+++ b/desloppify/languages/_framework/treesitter/imports/graph.py
@@ -31,14 +31,20 @@ def ts_build_dep_graph(
     query = _make_query(language, spec.import_query)
 
     scan_path = str(path.resolve())
-    file_set = set(file_list)
+    abs_file_list = [
+        filepath
+        if os.path.isabs(filepath)
+        else os.path.normpath(os.path.join(scan_path, filepath))
+        for filepath in file_list
+    ]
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks @tristanmanchester. I accepted the underlying problem but did not apply this PR as-is.

Decision: rejected in favor of the root #604 fix. The review found this PR was pointed at a real TypeScript graph path mismatch, but its normalization approach still had a subdirectory-scan condition: project-root-relative file entries could be joined against the scan path and duplicate path segments. I implemented the root-safe version in commit a9289015 by splitting source-root relative import resolution from tsconfig-root alias resolution.

Validation passed:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/languages/typescript/tests/test_ts_deps.py -q

Closing this PR because the accepted graph bug is fixed locally via #604.

peteromallet:
Closing after fixing the accepted graph-root issue locally in commit a9289015. Thanks again for the contribution.

PR #574, opened 2026-04-17
Handle validator exceptions as batch failures

tristanmanchester:
## Summary
- wrap runner output validation behind a safe helper that logs validator exceptions and treats them as invalid output
- use the guarded validator path in the initial validation, grace-wait retry loop, and stdout/stderr fallback recovery path
- add regressions for both the success helper and the public `execute_batches()` contract when a custom validator raises

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/commands/review/test_review_runner_batch_split_direct.py desloppify/tests/commands/review/test_review_runner_helpers_direct.py desloppify/tests/review/test_runner_parallel_internals.py`

### DIFF
--- a/desloppify/app/commands/review/runner_process_impl/attempt_success.py
+++ b/desloppify/app/commands/review/runner_process_impl/attempt_success.py
@@ -29,7 +29,7 @@ def handle_successful_attempt_core(
         return None
 
     validate = deps.validate_output_fn or default_validate_fn
-    valid = validate(output_file)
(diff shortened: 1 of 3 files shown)


peteromallet:

I applied the runner validation change as reviewed. Focused review-runner tests passed, and the full suite passed. Local commit: `e7b56d5d`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as e7b56d5d.

PR #577, opened 2026-04-17
Fix empty if-chain same-line else removal

tristanmanchester:
## Summary
- keep scanning `} else {` continuations in `_find_if_chain_end()` so brace depth stays balanced across the full if/else chain
- replace the premature `break` paths with `continue`, allowing the scanner to process the `{` that opens the else block
- add a regression proving the empty-if-chain fixer removes the whole same-line else chain instead of leaving orphaned braces behind

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/lang/typescript/test_ts_empty_if_chain_fixer_direct.py desloppify/tests/lang/typescript/test_typescript_detectors_fixers_and_wrappers_split_direct.py`

### DIFF
--- a/desloppify/languages/typescript/fixers/if_chain.py
+++ b/desloppify/languages/typescript/fixers/if_chain.py
@@ -57,12 +57,12 @@ def _find_if_chain_end(lines: list[str], start: int) -> int:
                 if found_brace and brace_depth == 0:
                     rest = line[ci + 1 :].strip()
                     if rest.startswith("else"):
-                        break
+                        continue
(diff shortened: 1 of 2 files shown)


peteromallet:

I applied the same-line else-chain fix as reviewed. The focused TypeScript fixer test passed, and the full project plus TypeScript language tests passed. Local commit: `596b1ed7`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as 596b1ed7.

PR #579, opened 2026-04-17
Guard move destinations against overwrite

tristanmanchester:
## Summary
- add an apply-time destination existence check immediately before filesystem moves
- prevent file and directory moves from overwriting destinations created after the command planning phase
- add regressions proving existing destinations are preserved for both move paths

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/commands/test_transitive_engine.py -k "ApplyFileMove or ApplyDirectoryMove or rollback_move_target"`

### DIFF
--- a/desloppify/app/commands/move/apply.py
+++ b/desloppify/app/commands/move/apply.py
@@ -17,6 +17,11 @@ from desloppify.base.output.fallbacks import restore_files_best_effort, warn_bes
 from desloppify.base.output.terminal import colorize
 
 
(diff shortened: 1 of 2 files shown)


peteromallet:

I applied the destination overwrite guard as reviewed. Focused move tests passed, and the full suite passed. Local commit: `2b11d259`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as 2b11d259.

PR #582, opened 2026-04-17
Preserve hidden path prefixes in exclusions

tristanmanchester:
## Summary
- replace the exclusion helper's broad `lstrip` normalization with exact `./` prefix removal
- preserve leading dots for hidden paths so `.cache/**` and `cache/**` remain distinct
- add direct regression coverage for hidden-directory glob matching

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/core/test_utils.py -k "matches_exclusion"`


peteromallet:

I applied the hidden-path exclusion fix as reviewed. Focused utility tests passed, and the full suite passed. Local commit: `b2fbc019`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as b2fbc019.

PR #583, opened 2026-04-17
Use live state for organize enrichment checks

tristanmanchester:
## Summary
- pass live state into organize-stage enrichment validation and organize confirmation
- prevent clusters containing only now-closed issues from blocking organize and confirm flows
- add direct regressions proving both blocking paths forward state to the enrichment gate

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/commands/plan/test_triage_split_modules_direct.py -k "organize_submission_passes_state_to_enrichment_gate or confirm_organize_passes_state_to_enrichment_gate or confirmation_modules_stage_presence_guards"`

### DIFF
--- a/desloppify/app/commands/plan/triage/confirmations/organize.py
+++ b/desloppify/app/commands/plan/triage/confirmations/organize.py
@@ -28,10 +28,10 @@ from ..validation.enrich_checks import (
 )
 
 
-def _require_enriched_clusters(plan: dict) -> bool:
+def _require_enriched_clusters(plan: dict, state: dict | None = None) -> bool:
     from ..stages.helpers import unenriched_clusters  # noqa: PLC0415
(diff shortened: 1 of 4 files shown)


peteromallet:

Decision: implemented. The review confirmed this was a real organize-stage state bug: the enrichment gates were using the plan without the live state, so closed review issues could still block organize validation/confirmation. I applied the narrow fix from this PR, added regressions for both paths, and committed it as b9736b02.

Validation passed:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/tests/commands/plan/test_triage_split_modules_direct.py -q

peteromallet:
Closing after applying the accepted fix locally in commit b9736b02. Thanks again for the contribution.

PR #586, opened 2026-04-17
Handle Rust strings in shared brace matching

tristanmanchester:
## Summary
- make the shared Rust detector brace and delimiter helpers ignore strings, raw strings, char literals, and comments
- reuse the existing string-aware delimiter matcher from the Rust tooling layer instead of maintaining a second naive implementation
- add regressions for public function extraction when strings and raw strings contain braces

## Testing
- `uv run --with pytest python -m pytest desloppify/languages/rust/tests/test_shared_direct.py`

### DIFF
--- a/desloppify/languages/rust/detectors/_shared.py
+++ b/desloppify/languages/rust/detectors/_shared.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from typing import Any
 
 from desloppify.base.discovery.file_paths import rel, resolve_path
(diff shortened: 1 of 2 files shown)


peteromallet:

I applied the Rust shared brace-matching change as reviewed. Focused Rust tests passed, and the full project plus Rust language test run passed. Local commit: `5f65a417`.

Closing this PR as applied.

peteromallet:
Closing after accepted review decision; applied locally as 5f65a417.

PR #587, opened 2026-04-17
Use full lines for Python relative import moves

tristanmanchester:
## Summary
- replace full relative import lines instead of generic `from ..` prefixes when updating Python move imports
- cover both importer rewrites and self-rewrites in the Python move helper
- add a regression proving `from .. import X` no longer corrupts sibling imports like `from ..utils import Y` during apply

## Testing
- `uv run --with pytest python -m pytest desloppify/languages/python/tests/test_py_move.py`

### DIFF
--- a/desloppify/languages/python/move.py
+++ b/desloppify/languages/python/move.py
@@ -49,6 +49,19 @@ def _replace_exact_module(line: str, old_module: str, new_module: str) -> str:
     return re.sub(rf"(?<!\w){re.escape(old_module)}(?![\w.])", new_module, line)
 
 
+def _replace_relative_from_module(
+    line: str,
+    old_module: str,
+    new_module: str,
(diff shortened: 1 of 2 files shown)


peteromallet:

Decision: implemented. The review confirmed this was a real Python move-command corruption bug: replacing a broad relative prefix such as `from ..` could accidentally rewrite sibling imports like `from ..utils import helper`. I applied the narrow fix so relative from-import moves replace the matched full import line/module portion, added the regression from the PR, and committed it as 1860dd64.

Validation passed:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/languages/python/tests/test_py_move.py -q

peteromallet:
Closing after applying the accepted fix locally in commit 1860dd64. Thanks again for the contribution.

PR #588, opened 2026-04-17
Mark all emptied resolve clusters done

tristanmanchester:
## Summary
- track all clusters affected by a multi-pattern resolve instead of only the first matching cluster
- mark every cluster emptied by the resolve as `done` and emit a `cluster_done` log entry for each
- add regressions for multi-cluster completion in the resolve living-plan path

## Testing
- `uv run --with pytest python -m pytest desloppify/tests/commands/resolve/test_living_plan_direct.py`

### DIFF
--- a/desloppify/app/commands/resolve/living_plan.py
+++ b/desloppify/app/commands/resolve/living_plan.py
@@ -45,16 +45,41 @@ class ClusterContext(NamedTuple):
     cluster_remaining: int
 
 
+def _affected_cluster_names(plan: dict, resolved_ids: list[str]) -> list[str]:
+    """Return unique cluster names referenced by the resolved ids."""
+    overrides = plan.get("overrides") or {}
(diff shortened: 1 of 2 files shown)


peteromallet:

Decision: implemented. The review confirmed this was a real living-plan resolve bug: resolving items across multiple clusters could empty several clusters while only the first affected cluster was marked done. I applied the fix so all affected clusters are captured before purge, every fully emptied cluster gets `execution_status: done` plus a `cluster_done` log entry, and active focus is cleared when appropriate. Committed as f8cf0417.

Validation passed:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/tests/commands/resolve/test_living_plan_direct.py -q

peteromallet:
Closing after applying the accepted fix locally in commit f8cf0417. Thanks again for the contribution.

PR #589, opened 2026-04-17
Preserve else chains in TS log cleanup

tristanmanchester:
## Summary
- keep empty `if` and `else if` blocks in place when an `else` continuation follows on the next non-blank line
- prevent `remove_empty_blocks()` from orphaning `else` and `else if` clauses after debug-log removal
- add regressions that run the real TypeScript log fixer and validate the output with `node --check`

## Testing
- `uv run --with pytest python -m pytest desloppify/languages/typescript/tests/test_ts_fixers.py -k "orphaned_else or fix_debug_logs"`

### DIFF
--- a/desloppify/languages/typescript/fixers/logs_cleanup.py
+++ b/desloppify/languages/typescript/fixers/logs_cleanup.py
@@ -200,6 +200,13 @@ def _try_remove_multiline_block(
         new_lines.append(f"{indent}}}\n")
         return j + 1
     if re.match(r"\s*(?:if|else\s+if)\s*\(", stripped):
+        next_line_idx = j + 1
+        while next_line_idx < len(lines) and lines[next_line_idx].strip() == "":
+            next_line_idx += 1
+        if next_line_idx < len(lines):
+            next_stripped = lines[next_line_idx].strip()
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks for the PR. I accepted the TypeScript log-cleanup bug: removing log-only `if` / `else if` blocks could leave an orphaned `else`.

Decision: implemented with changes in `3cb04149`. The production fix preserves empty `if` / `else if` blocks when the next nonblank line continues an `else` chain. I changed the regression approach so it asserts the fixer output structurally instead of relying on `node --check` accepting `.ts` files, which was not portable on the local Node version.

Validation run:
- `PYENV_VERSION=3.11.11 python -m pytest desloppify/tests/review/policy/test_review_dimensions_direct.py desloppify/tests/core/test_config_schema_direct.py desloppify/tests/lang/common/test_bash_unused_imports.py desloppify/languages/typescript/tests/test_ts_fixers.py -q`

Closing this PR as superseded by the local implementation.

peteromallet:
Closed as superseded by local commit `3cb04149`, which keeps the accepted behavior and uses portable regression tests.

peteromallet:

PR #596, opened 2026-04-19
fix: propagate ContextVar to prefetch executor workers

Wolframko:
## Problem

Running a scan on a monorepo with large vendored trees under `tmp/`, `.refs/`, and feature worktrees reliably produced:

```
WARNING: Boilerplate duplication detection skipped: jscpd timed out after 120s.
```

…even after the user declared those directories in `config.exclude` via `desloppify exclude`. The same mechanism also silently mis-scopes the security prefetch for any language plugin that uses `lang.detect_lang_security_detailed` through the shared prefetch path.

## Root cause

`_PREFETCH_EXECUTOR` in `desloppify/languages/_framework/base/shared_phases_review.py` is a module-level `concurrent.futures.ThreadPoolExecutor(max_workers=2)` used for jscpd and language-security prefetch tasks.

Python's `ThreadPoolExecutor` does **not** propagate `ContextVar` values into worker threads. Several pieces of runtime state live in a `ContextVar` — most importantly the exclusion patterns set by `_apply_persisted_exclusions` → `set_exclusions(...)` in `cli.py`, which are read back via `get_exclusions()` when `_jscpd_ignore_arg` builds jscpd's `--ignore` list.

Inside the pool worker, the `ContextVar` lookup returned its default (`None`), `current_runtime_context()` fell back to the process-default `_PROCESS_RUNTIME_CONTEXT`, and `get_exclusions()` returned `()`. `_jscpd_ignore_arg` then degraded to `_BASE_IGNORES` only, so jscpd happily walked into `tmp/screenpipe/**`, `.refs/**`, worktree copies, etc. — exactly the directories the user excluded — and burned through its 120 s budget.

I verified the degradation by printing `_jscpd_ignore_arg(Path('.'))` from (a) the CLI context and (b) a `_PREFETCH_EXECUTOR.submit` worker: the worker's ignore list was strictly the `_BASE_IGNORES` set, while the CLI call included all of the user's persisted excludes.

## Fix

Add `_submit_with_context(fn, *args, **kwargs)` that wraps the submitted callable in `contextvars.copy_context().run(...)` before passing it to `_PREFETCH_EXECUTOR.submit(...)`. Route both prefetch submissions (jscpd boilerplate detection and language-security prefetch) through this helper.

The copied context carries the ContextVar bindings present at submit time, so `get_exclusions()` in the worker now sees the per-scan runtime context and `_jscpd_ignore_arg` emits the full ignore list.

## Tests

Added `desloppify/tests/lang/common/test_shared_phases_prefetch_context.py`:

- `test_submit_with_context_preserves_runtime_exclusions` — sets exclusions in a `runtime_scope`, submits a read-back callable through `_submit_with_context`, asserts the worker observes the same tuple. Fails on `main`, passes with the fix.
- `test_submit_with_context_isolates_between_scopes` — two sequential `runtime_scope` blocks with different exclusions must produce different worker-side snapshots (no cross-scope bleed).

Full suite (5532 passed, 153 skipped, 12 deselected) still green. The one deselected failure — `test_no_extra_bundled_files` complaining about `scoring.md` — predates this PR and is unrelated.

## Repro

On a repo with a big directory that is in `config.exclude` but not in `_BASE_IGNORES` (e.g. `tmp/` pointing at 200+ MB of vendored code):

```bash
desloppify exclude tmp
desloppify scan --path .
# → "WARNING: Boilerplate duplication detection skipped: jscpd timed out after 120s."
```

Inspect the running jscpd argv with `ps -eww`:

- Before: `--ignore` lists only `_BASE_IGNORES` (`**/.git/**`, `**/node_modules/**`, …), no `**/tmp/**`.
- After: `--ignore` includes every user-configured exclude basename.

### DIFF
--- a/desloppify/languages/_framework/base/shared_phases_review.py
+++ b/desloppify/languages/_framework/base/shared_phases_review.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import concurrent.futures
+import contextvars
 import hashlib
 import logging
 import os
@@ -50,6 +51,20 @@ _FUNCTION_CACHE_ATTR = "_shared_review_function_cache"
 _PREFETCH_BOILERPLATE_KEY = "boilerplate"
 _PREFETCH_SECURITY_KEY = "security_lang"
 _PREFETCH_EXECUTOR = concurrent.futures.ThreadPoolExecutor(max_workers=2)
+
+
+def _submit_with_context(fn: Callable[..., Any], *args: Any, **kwargs: Any) -> concurrent.futures.Future[Any]:
+    """Submit *fn* to the prefetch executor, preserving the caller's contextvars.
+
+    ``ThreadPoolExecutor`` does not propagate ``ContextVar`` values to worker
+    threads. Several runtime settings (notably user-configured path exclusions
+    read via ``get_exclusions()``) live in a ContextVar and are only observable
+    from the calling thread. Without this wrapper, detectors launched in the
+    pool (jscpd, security prefetch) see the process-default context and ignore
+    the current scan's configuration.
+    """
+    ctx = contextvars.copy_context()
+    return _PREFETCH_EXECUTOR.submit(ctx.run, fn, *args, **kwargs)
 
 
 def _detector_cache(review_cache: object, detector: str) -> dict[str, object] | None:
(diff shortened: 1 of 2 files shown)


pararti:
@claude close it

peteromallet:
Thanks @Wolframko. I accepted and applied this change.

Decision: implemented. The review confirmed this was a real runtime-context propagation bug: jscpd and language-security prefetch work run in `ThreadPoolExecutor` workers, which do not inherit the scan ContextVars by default. That meant configured exclusions could be visible to the CLI thread but missing inside prefetch workers. I applied the context-preserving submission helper, routed both prefetch paths through it, added the runtime exclusion propagation/isolation regressions, and committed it as 65a7fd6c.

Validation passed:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/tests/lang/common/test_shared_phases_prefetch_context.py -q

peteromallet:
Closing after applying the accepted fix locally in commit 65a7fd6c. Thanks again for the contribution.

PR #598, opened 2026-04-23
fix: resolve relative/absolute path mismatch in ts_build_dep_graph + add Ruby test coverage module

wbsmolen:
## Problem

`ts_build_dep_graph` silently discards all import edges when `file_set` uses relative paths.

**Root cause**: `find_source_files` returns relative paths, so `file_set = set(file_list)` contains entries like `scripts/asc_client.rb`. But the function then normalizes `resolved` to an absolute path via `os.path.normpath(os.path.join(scan_path, resolved))`, making the membership check `if resolved not in file_set` always True — so no edge is ever added.

This affects all language plugins that use tree-sitter dep graph building (TypeScript, PHP, Ruby, etc.) whenever `find_source_files` is used as the file finder.

**Reproduction**: Point desloppify at a Ruby project with `require_relative` imports. The dep graph will always be empty `{}` despite correct tree-sitter parsing.

## Fix

Try the resolved path as-is first, then try the alternative form (absolute ↔ relative) before discarding. This handles both:
- Relative `file_set` + relative `resolved` (most backends via `find_source_files`)
- Absolute `file_set` + relative `resolved` (existing test fixtures)

## Ruby test coverage module

Also adds `desloppify/languages/ruby/test_coverage.py` with the required hooks:
- `parse_test_import_specs`: extracts `require`/`require_relative` paths
- `resolve_import_spec`: resolves relative paths from spec files to production files
- `has_testable_logic`: detects `def`/`class`/`module` definitions
- `strip_test_markers`: strips `_spec.rb`, `test_*.rb`, `*_test.rb` suffixes
- `strip_comments`: strips `#` comments while preserving string literals

Without this module, `get_lang_hook("ruby", "test_coverage")` returned `None`, leaving all Ruby scripts marked as untested regardless of the dep graph state.

And wires it in via `generic_lang(..., test_coverage_module=_test_coverage_mod)` in `ruby/__init__.py`.

## Tests

5543 passed, 153 skipped (1 pre-existing failure in `test_bundled_sync` unrelated to this change).

### DIFF
--- a/desloppify/languages/_framework/treesitter/imports/graph.py
+++ b/desloppify/languages/_framework/treesitter/imports/graph.py
@@ -75,13 +75,23 @@ def ts_build_dep_graph(
             if resolved is None:
                 continue
 
-            # Normalize to absolute path.
-            if not os.path.isabs(resolved):
-                resolved = os.path.normpath(os.path.join(scan_path, resolved))
-
-            # Only track edges within the scanned file set.
+            # file_set may use relative or absolute paths depending on the
+            # file_finder. Try the resolved path as-is first, then try the
+            # absolute form (for relative resolvers with an absolute file_set)
+            # and the relative form (for absolute resolvers with a relative
+            # file_set).
             if resolved not in file_set:
-                continue
+                if os.path.isabs(resolved):
+                    try:
+                        alt = os.path.relpath(resolved, scan_path)
+                    except ValueError:
+                        alt = resolved
+                else:
+                    alt = os.path.normpath(os.path.join(scan_path, resolved))
+                if alt in file_set:
+                    resolved = alt
+                else:
+                    continue
 
             graph[filepath]["imports"].add(resolved)
             if resolved in graph:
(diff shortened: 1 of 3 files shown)


peteromallet:
Thanks @wbsmolen. I accepted the underlying TypeScript graph problem but did not apply this PR as-is.

Decision: rejected in favor of the root #604 fix. The graph portion overlapped with #572/#604 and is now fixed locally in commit a9289015 with explicit source-root vs tsconfig-root handling. I did not apply the Ruby coverage portion because the Stage 1 conditions remain unresolved: it needs regression tests and broader `require "gem/file"` -> `lib/gem/file.rb` handling before it should land.

Validation passed for the graph fix:
- PYENV_VERSION=3.11.11 python -m pytest desloppify/languages/typescript/tests/test_ts_deps.py -q

Closing this PR because the accepted graph bug is fixed locally and the Ruby portion needs a follow-up with the requested tests/behavior.

peteromallet:
Closing after fixing the accepted TypeScript graph issue locally in commit a9289015. The Ruby coverage part still needs the Stage 1 conditions before landing. Thanks again for the contribution.

PR #600, opened 2026-04-27
fix: expand bandit --exclude paths into globs so nested files are pruned

f0rodo:
## Problem

The Security dimension was reporting 5,958 issues on a project that had `.claude` and `.claude/worktrees` in its `exclude` config. Inspection showed 5,932 of those were in `.claude/worktrees/...` — exactly the directories the user told desloppify to skip.

The exclusion is honored by the file-discovery layer (`_find_source_files_cached` correctly prunes via `_is_excluded_dir`). But the `bandit_adapter` calls bandit with the project root as its scan target and a `--exclude` value built from `collect_exclude_dirs()` — a list of bare absolute paths like `/repo/.claude`.

Bandit's `--exclude` is matched against the filenames it discovers during its own recursive walk: it filters files whose **name equals** one of the supplied strings. So `/repo/.claude` filters the directory entry itself, but `/repo/.claude/worktrees/foo.py` is still scanned (the full filename does not equal the exclude string).

Repro:
```bash
# Project with .claude/worktrees containing test files (assert-heavy)
$ bandit -r --exclude /repo/.claude /repo
# Returns thousands of B101 findings inside .claude/worktrees/
```

vs the form bandit's pattern matcher actually understands:
```bash
$ bandit -r --exclude '**/.claude/**' /repo
# Properly prunes the subtree
```

## Fix

`_exclude_globs(paths)` (new helper in `bandit_adapter.py`) walks the path components of each input and emits `**/<name>` + `**/<name>/**` glob siblings alongside the original absolute path. The originals are kept as a belt-and-braces fallback; the globs are what actually do the work.

`detect_with_bandit` now passes `_exclude_globs(exclude_dirs)` to `--exclude` instead of the raw list.

## Verified

End-to-end on a real project that excludes `.claude` and `.claude/worktrees`:

| | Before | After |
|---|---|---|
| Security findings | 5,958 | 3 |
| In `.claude/worktrees/...` | 5,932 | 0 |
| Security score | 8.9% | 99.8% |

## Test

`test_exclude_dirs_expanded_to_glob_for_nested_pruning` asserts the emitted `--exclude` arg contains `**/.claude`, `**/.claude/**`, and `**/worktrees/**` when given `/project/.claude` and `/project/.claude/worktrees` as input.

5,533 existing tests still pass; the unrelated `test_bundled_sync` failure on `scoring.md` predates this branch.

### DIFF
--- a/desloppify/languages/python/detectors/bandit_adapter.py
+++ b/desloppify/languages/python/detectors/bandit_adapter.py
@@ -185,6 +185,43 @@ def _to_security_entry(
     }
 
 
+def _exclude_globs(exclude_dirs: list[str]) -> list[str]:
+    """Expand absolute exclude paths into bandit-friendly glob patterns.
+
+    Bandit's ``--exclude`` is matched against filenames it discovers during
+    its recursive walk. When desloppify hands it absolute directory paths
+    (e.g. ``/repo/.claude``), bandit only filters files whose name *equals*
+    one of those paths — directories *under* the exclude (e.g.
+    ``/repo/.claude/worktrees/foo.py``) still get scanned because their
+    full filename does not equal the exclude string.
+
+    Expanding each absolute path into ``**/<basename>/**`` (and a sibling
+    ``<basename>`` form for the directory itself) ensures bandit prunes the
+    whole subtree the way the rest of desloppify already does. The
+    original absolute paths are kept as a belt-and-braces fallback.
+    """
+    expanded: list[str] = []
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks @f0rodo for the contribution/report. Final decision: reject. The Bandit exclude fix is based on a false premise and expands absolute paths into broad globs that can accidentally exclude the scan root or common parent directories.

peteromallet:
Closing after review decision.

PR #602, opened 2026-05-08
fix(bash): filter unused-import detector to source / . only

raveinid:
## Problem

`BASH_SPEC.import_query` in `desloppify/languages/_framework/treesitter/specs/scripting.py` captures every `(command, argument)` pair in any bash file as an "import". The unused-import phase then runs on those raw captures, so every shell flag becomes a spurious finding:

| Shell line | Reported as |
|---|---|
| `set -euo pipefail` | Unused import: `pipefail` |
| `curl -fsS https://...` | Unused import: `-fsS` |
| `find . -name '*.tmp'` | Unused import: `-name` |
| `cut -d: -f2` | Unused import: `-d` |

`resolve_bash_source` only resolves `source`/`.` invocations, but the unused-import phase flags on the raw captured `path` *before* resolution, so the resolver can't filter them out.

On a small bash-heavy repo (8 shell scripts) this produced ~80 false-positive `unused::*::unused_import` findings on a clean codebase.

## Fix

Add a `#match?` predicate **inside** the `(command ...)` pattern, restricting the query to `source <path>` and `. <path>`. Mirrors the existing `(#eq? @_fn "include")` pattern Julia's spec uses for `include(...)` calls in `functional.py`.

```diff
 import_query="""
     (command
         name: (command_name) @_cmd
+        (#match? @_cmd "^(source|\\\\.)$")
         argument: (word) @path) @import
 """,
```

Note: the predicate must be placed **inside** the pattern parens — placed outside it parses as a separate top-level pattern with no captures and has no filtering effect.

## Tests

New file `desloppify/tests/lang/common/test_bash_unused_imports.py` — 4 regression tests covering both halves of the contract:

- shell flags (`set -euo`, `curl -fsS`, `find -name`, `cut -d:`) → 0 findings
- unused `source ./helpers.sh` → flagged
- unused `. ./extras.sh` (POSIX form) → flagged
- used `source ./helpers.sh` referenced later → not flagged

```
$ pytest desloppify/tests/lang/common/test_treesitter.py \
         desloppify/tests/lang/common/test_treesitter_complexity_and_integration.py \
         desloppify/tests/lang/common/test_bash_unused_imports.py -q
151 passed, 1 skipped in 1.67s
```

## Verification on a real repo

Same scan target before and after the fix:

|  | Before | After |
|---|---|---|
| `unused::*::unused_import` findings on bash files | ~80 | 0 |
| Coverage of real `source` directives | drowned in noise | preserved |

### DIFF
--- a/desloppify/languages/_framework/treesitter/specs/scripting.py
+++ b/desloppify/languages/_framework/treesitter/specs/scripting.py
@@ -47,9 +47,22 @@ BASH_SPEC = TreeSitterLangSpec(
             body: (compound_statement) @body) @func
     """,
     comment_node_types=frozenset({"comment"}),
+    # Filter to `source <path>` and `. <path>` only. The #match? predicate
+    # must be placed INSIDE the (command ...) pattern (the Julia spec in
+    # functional.py uses #eq? the same way for `include(...)`); placed
+    # outside the closing paren it parses as a separate top-level pattern
+    # and has no filtering effect.
+    #
+    # Without the predicate, the query matched every (command, argument)
+    # pair, so `set -euo pipefail` produced "Unused import: pipefail" and
+    # `curl -fsS https://...` produced "Unused import: -fsS" — every shell
+    # flag was a spurious finding. resolve_bash_source only resolves
+    # source/. invocations, but unused-imports flagging runs on the raw
+    # captured `path` before resolution, so the resolver couldn't save it.
     import_query="""
         (command
             name: (command_name) @_cmd
+            (#match? @_cmd "^(source|\\.)$")
             argument: (word) @path) @import
     """,
     resolve_import=resolve_bash_source,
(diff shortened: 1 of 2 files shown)


peteromallet:
Thanks for the PR. I accepted the Bash unused-import bug: the detector was treating ordinary command arguments and flags as imports.

Decision: implemented with changes in `e8f0ec39`. The import query now only captures `source` and `.` commands, and it is constrained to the first argument after the command so calls like `source ./helpers.sh foo bar` do not treat `foo` or `bar` as imports. I added regressions for shell flags, source/dot-source imports, used sources, and extra source arguments.

Validation run:
- `PYENV_VERSION=3.11.11 python -m pytest desloppify/tests/review/policy/test_review_dimensions_direct.py desloppify/tests/core/test_config_schema_direct.py desloppify/tests/lang/common/test_bash_unused_imports.py desloppify/languages/typescript/tests/test_ts_fixers.py -q`

Closing this PR as superseded by the local implementation.

peteromallet:
Closed as superseded by local commit `e8f0ec39`, which keeps the accepted fix and adds the first-argument condition requested in review.

peteromallet:

PR #605, opened 2026-05-10
fix: cap incompatible tree-sitter language pack

GDownes:
## Problem
Running the latest GitHub install against a TypeScript project can crash in the tree-sitter cohesion phase when pip resolves `tree-sitter-language-pack` 1.8.0. The crash is `TypeError: __new__() argument 1 must be tree_sitter.Language, not builtins.Language`.

## Fix
Cap `tree-sitter-language-pack` below 1.8 in the `treesitter` and `full` extras, and add a packaging regression test for the cap. I verified that 1.6.2 returns `tree_sitter.Language` objects and that the TypeScript scan completes.

## Verification
- `python -m pytest desloppify/tests/core/test_pyproject_optional_dependencies.py -q` passes.
- `python -m desloppify --lang typescript scan --path /Users/gerarddownes/.codex/worktrees/f8b1/workorderguard/infra/core --state /tmp/desloppify-fix-verify-state.json --no-badge` completes without the crash.
- `python -m pytest desloppify/tests/ -q` has one unrelated existing failure: `test_no_extra_bundled_files` reports bundled `scoring.md` has no matching `docs/scoring.md`.

### DIFF
--- a/desloppify/tests/core/test_pyproject_optional_dependencies.py
+++ b/desloppify/tests/core/test_pyproject_optional_dependencies.py
@@ -49,3 +49,17 @@ def test_treesitter_extra_declares_runtime_and_language_pack() -> None:
     package_names = _package_names(treesitter_specs)
     assert "tree-sitter" in package_names
     assert "tree-sitter-language-pack" in package_names
+
+
+def test_treesitter_language_pack_is_capped_below_incompatible_release() -> None:
+    optional = _optional_dependencies()
+    treesitter_specs = optional.get("treesitter")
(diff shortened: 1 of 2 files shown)


harlequinetcie:
Tried this in local, and it works for me. Thanks for the fix!

peteromallet:
Thanks @GDownes for the fix. Final decision: merged with the reviewed approach.

I applied the dependency cap for `tree-sitter-language-pack<1.8` in both optional extras and kept the regression test that locks the packaging metadata. The change is committed locally as `c9db34b0`.

Closing this PR as accepted and applied.

peteromallet:
Closing after accepted review decision; applied locally as c9db34b0.

PR #612, opened 2026-05-20
fix(go): support golangci-lint v2 json output

masterkain:
## Summary

- update the Go plugin golangci-lint command to use golangci-lint v2 JSON output flags
- keep text/stats output disabled so the existing JSON parser receives clean JSON only
- add a focused regression test for the registered Go golangci-lint phase command

## Verification

- `.venv/bin/python -m pytest -q desloppify/languages/go/tests/test_init.py -k golangci` failed before the fix with `--out-format=json` still present
- `.venv/bin/python -m pytest -q desloppify/languages/go/tests`
- `.venv/bin/python -m pytest -q desloppify/tests/lang/common/test_generic_plugin.py`
- `.venv/bin/python -m ruff check desloppify/languages/go/__init__.py desloppify/languages/go/tests/test_init.py`
- `.venv/bin/desloppify scan --path <temporary-go-module> --no-badge` with local golangci-lint v2.9.0


masterkain:
not even a reply

Issue #418, opened 2026-03-13
Framework detection architecture: spec-driven horizontal layer (like tree-sitter)


peteromallet:
## Context

PR #414 adds Next.js framework detection — the first framework-specific detector in the codebase. The scanners and detection logic are solid, but before we establish the pattern that future frameworks will follow, we should think about what scales to 20 languages × 20 frameworks.

## Current plugin architecture

The codebase has two plugin depth levels and several horizontal layers:

```
                        Shallow                              Deep
                        (generic_lang)                       (LangConfig class)
                        ─────────────────────────────────────────────────
Language depth:         JS, Nim, Zig, Bash                   TS, Go, Rust, C++

Horizontal layers (work at any depth):
  Tree-sitter:          ✓ (via treesitter_spec=)             ✓ (via *all_treesitter_phases())
  Tool specs:           ✓ (via tools=[...])                  ✓ (via make_tool_phase())
```

Tree-sitter is the key precedent: it's spec-driven, generates `DetectorPhase` objects, and works with both shallow and deep plugins. Adding a new language's tree-sitter support = adding a spec, not modifying infrastructure.

## The question: what are frameworks?

Most framework-specific issues fall into four categories:

1. **File convention checks** — "error.tsx needs `use client`", "Django views belong in views.py". File pattern + content pattern. Mechanical.
2. **API misuse** — "`use server` in a client module", "useEffect only to sync state". String or AST pattern matches.
3. **External tool output** — `next lint`, `django check`. Tool running + output parsing.
4. **Architectural violations** — "mixing pages/ and app/ router", "circular model imports". Needs dep graph / cross-file analysis.

Categories 1-3 are shallow — pattern matching and tool running. Category 4 needs the language plugin's deep infrastructure (dep graph, zone map), but doesn't need deep *framework-specific* wiring — it just consumes what the language plugin already provides.

**Frameworks aren't a new depth level. They're a horizontal layer, like tree-sitter.**

## Proposed architecture

```
_framework/
  frameworks/
    types.py              # FrameworkSpec, ScannerRule, DetectionConfig
    registry.py           # registered framework specs
    detection.py          # generic "is this framework present?"
                          #   reads package.json, requirements.txt, Cargo.toml, go.mod
                          #   keyed by ecosystem
    phases.py             # framework_phases(lang_name) → list[DetectorPhase]
                          #   analogous to all_treesitter_phases()
    specs/
      nextjs.py           # Next.js: detection config + scanner rules + issue templates
[...]

PR #614, opened 2026-05-28
feat: Astro/Svelte/Vue project support via FrameworkSpec registry

elfensky:
## The problem

When you run `desloppify scan` on an Astro, Svelte, or Vue project, files that are imported only from `.astro`/`.svelte`/`.vue` get flagged as "orphaned" even though they are used.

On a real Astro 5 project ([descent-marketing](https://github.com/)), `desloppify scan --path .` reported **16 files as orphaned**. All 16 were genuinely imported — by `astro.config.mjs`, by `.astro` pages, by other `.js` modules.

Two separate things were broken:

1. **The JS plugin didn't know how to read `.astro`/`.svelte`/`.vue` files.** They aren't in its extension list, so it never opened them to see what they import. Any `.js` file imported only from a `.astro` page had 0 importers in the graph → flagged as orphan.

2. **The tree-sitter import pass silently dropped every import edge whenever the scan ran from a directory other than the project root.** `file_set` held relative paths and `resolve_import` produced absolute paths, so they never matched. Tests didn't notice because `desloppify scan` always chdir's to the project root in production.

## The solution

Two changes that work together.

### 1. Put framework file extensions on `FrameworkSpec` (#418)

The `FrameworkSpec` horizontal layer already exists for tool integrations and scanner rules — Next.js uses it today. Extend it with one new field:

```python
@dataclass(frozen=True)
class FrameworkSpec:
    ...
    source_extensions: tuple[str, ...] = ()  # NEW
```

Register three new specs in `_framework/frameworks/specs/`: `ASTRO_SPEC`, `SVELTE_SPEC`, `VUE_SPEC`. Each declares its own `source_extensions` (`(".astro",)`, `(".svelte",)`, `(".vue",)`).

Both the JS plugin and the TS plugin look the list up from the registry:

```python
framework_extensions = framework_source_extensions(ecosystem="node")
```

After this change, **adding a new framework like Qwik is one new file** (`specs/qwik.py`) plus a `register_framework_spec(...)` line. Zero edits to `graph.py`, `registration.py`, or the TS plugin.

### 2. Fix the cross-CWD bug in the dep-graph builder

Absolutize `file_list` at the top of `ts_build_dep_graph` so graph keys, tree-sitter file reads, and import-resolver outputs all share one coordinate system:

```python
absolute_file_list = [resolve_path(f) for f in file_list]
file_set = set(absolute_file_list)
```

Add an opt-in `framework_extensions` parameter that runs a regex pass over framework files and records them as importer edges. Framework files themselves are not added as graph nodes, so they never appear in orphan or coupling reports.

## Net effect on the Astro project

| | Before | After |
|---|---|---|
| Orphan findings | 16 | 0 |
| Import edges in JS graph | 0 | 56 |
| Importer edges in JS graph | 0 | 60 |

All 16 "orphans" were real importers; they now correctly show importers from `astro.config.mjs`, `Starfield.astro`, `WaitlistForm.astro`, `index.astro`, and the internal `.js → .js` edges that the cross-CWD bug had also been hiding.

## What changes per file

- `_framework/frameworks/types.py` — adds `source_extensions` field to `FrameworkSpec`
- `_framework/frameworks/registry.py` — adds `framework_source_extensions(ecosystem)` helper; registers the 3 new specs
- `_framework/frameworks/specs/{astro,svelte,vue}.py` — new files, one per framework
- `_framework/treesitter/imports/graph.py` — absolutizes paths; adds the `framework_extensions` regex pass
- `_framework/generic_support/{core,registration}.py` — wires `frameworks=True` through to the registry query
- `typescript/detectors/deps/__init__.py` — replaces the hand-rolled `_FRAMEWORK_EXTENSIONS` constant with the same registry query at both call sites (line 88 framework-file pass, line 218 dynamic-imports helper)

## Tests

Adds 9 tests, modifies 1 existing monkeypatch lambda:

- `tests/lang/common/test_treesitter_imports_direct.py`
  - `test_graph_records_framework_files_as_importers` — exercises `framework_extensions` directly; asserts framework files are importers but **not** graph nodes.
  - `test_graph_framework_pass_is_opt_in` — guards default-off behaviour so callers without `frameworks=True` are unaffected.
- `javascript/tests/test_init.py`
  - `test_dep_graph_treats_astro_frontmatter_as_importer` — end-to-end via the registered JS plugin.
  - `test_dep_graph_treats_mjs_config_as_importer` — covers the cross-CWD `.mjs → .js` regression.
- `tests/lang/common/test_framework_source_extensions.py` (new file, 5 tests)
  - Built-in node-ecosystem defaults include `.astro` / `.svelte` / `.vue`
  - Sort + deduplication semantics
  - Specs without `source_extensions` contribute nothing (NEXTJS_SPEC stays unaffected)
  - Ecosystem filtering excludes other ecosystems
  - Unfiltered query aggregates across ecosystems

Suite: **6829 passed, 19 skipped**. No regressions.

## CI unblock (separate commit)

The second commit on this PR fixes 5 pre-existing test failures on `main` so this PR can actually go green. Those failures have been red on every PR since 2026-04-06 and are mechanically simple — they don't belong to this feature. Details in the comment thread.

## What this does NOT do

- **Migrate the TS plugin's hand-rolled `build_dep_graph`** onto the shared `ts_build_dep_graph`. That's a real refactor — it needs lifting `external_imports` and tsconfig-alias resolution into the `TreeSitterLangSpec` layer first.
- **Resolve the graph-node semantics divergence** between the TS plugin (treats framework files as graph nodes) and the JS plugin (doesn't).
- **Split the overloaded `frameworks: bool` flag** into `framework_phases` + `framework_imports`.

These are intentional follow-ups, not in scope here.

Related architecture issue: #418.

### DIFF
--- a/desloppify/languages/_framework/frameworks/registry.py
+++ b/desloppify/languages/_framework/frameworks/registry.py
@@ -8,6 +8,18 @@ from .types import FrameworkSpec
 
 FRAMEWORK_SPECS: dict[str, FrameworkSpec] = {}
 
+# Per-ecosystem cache for ``framework_source_extensions``. Invalidated by any
+# registry mutation (``register_framework_spec``). Keyed by the normalized
+# ecosystem string ("" for the unfiltered query) so each call shape gets its
+# own slot. In production the registry is immutable after startup, so this
+# caches a single tuple per ecosystem for the process lifetime; in tests the
+# fixture-driven mutations invalidate it cleanly.
+_EXTENSIONS_CACHE: dict[str, tuple[str, ...]] = {}
+
+
+def _invalidate_extensions_cache() -> None:
+    _EXTENSIONS_CACHE.clear()
+
 
 def register_framework_spec(spec: FrameworkSpec) -> None:
     """Register a framework spec by id."""
@@ -15,6 +27,7 @@ def register_framework_spec(spec: FrameworkSpec) -> None:
     if not key:
         raise ValueError("FrameworkSpec.id must be non-empty")
     FRAMEWORK_SPECS[key] = spec
+    _invalidate_extensions_cache()
 
 
 def get_framework_spec(framework_id: str) -> FrameworkSpec | None:
@@ -63,9 +76,16 @@ def framework_source_extensions(ecosystem: str | None = None) -> tuple[str, ...]
     of the host language's modules. Lookup is driven by the registry, so
     adding a new framework with ``source_extensions`` automatically extends
     every plugin that calls this — no infrastructure edits required.
+
+    Cached per ecosystem key; ``register_framework_spec`` invalidates the
+    cache so tests that mutate the registry see fresh results.
     """
     ensure_builtin_specs_loaded()
-    return tuple(
+    cache_key = (ecosystem or "").strip().lower()
+    cached = _EXTENSIONS_CACHE.get(cache_key)
+    if cached is not None:
+        return cached
+    result = tuple(
         sorted(
             {
                 ext
@@ -74,6 +94,8 @@ def framework_source_extensions(ecosystem: str | None = None) -> tuple[str, ...]
             }
         )
     )
+    _EXTENSIONS_CACHE[cache_key] = result
+    return result
 
 
 __all__ = [
--- a/desloppify/languages/_framework/frameworks/registry.py
+++ b/desloppify/languages/_framework/frameworks/registry.py
@@ -39,9 +39,15 @@ def _register_builtin_specs() -> None:
     """Register built-in framework specs shipped with the repo."""
     if FRAMEWORK_SPECS:
         return
+    from .specs.astro import ASTRO_SPEC
     from .specs.nextjs import NEXTJS_SPEC
+    from .specs.svelte import SVELTE_SPEC
+    from .specs.vue import VUE_SPEC
 
     register_framework_spec(NEXTJS_SPEC)
+    register_framework_spec(ASTRO_SPEC)
+    register_framework_spec(SVELTE_SPEC)
+    register_framework_spec(VUE_SPEC)
 
 
 def ensure_builtin_specs_loaded() -> None:
@@ -49,9 +55,31 @@ def ensure_builtin_specs_loaded() -> None:
     _register_builtin_specs()
 
 
+def framework_source_extensions(ecosystem: str | None = None) -> tuple[str, ...]:
+    """Return the sorted, deduplicated source extensions across registered specs.
+
+    Used by dep-graph builders to learn which non-host-language file types
+    (e.g. ``.astro``, ``.svelte``, ``.vue``) should be scanned as importers
+    of the host language's modules. Lookup is driven by the registry, so
+    adding a new framework with ``source_extensions`` automatically extends
+    every plugin that calls this — no infrastructure edits required.
+    """
+    ensure_builtin_specs_loaded()
+    return tuple(
+        sorted(
+            {
+                ext
+                for spec in list_framework_specs(ecosystem=ecosystem).values()
+                for ext in spec.source_extensions
+            }
+        )
+    )
+
+
 __all__ = [
     "FRAMEWORK_SPECS",
     "ensure_builtin_specs_loaded",
+    "framework_source_extensions",
     "get_framework_spec",
     "list_framework_specs",
     "register_framework_spec",
--- a/desloppify/languages/_framework/frameworks/specs/astro.py
+++ b/desloppify/languages/_framework/frameworks/specs/astro.py
@@ -1,0 +1,25 @@
+"""Astro framework spec (Node ecosystem)."""
+
+from __future__ import annotations
+
+from ..types import DetectionConfig, FrameworkSpec
+
+ASTRO_SPEC = FrameworkSpec(
+    id="astro",
+    label="Astro",
+    ecosystem="node",
+    detection=DetectionConfig(
+        dependencies=("astro",),
+        config_files=(
+            "astro.config.mjs",
+            "astro.config.js",
+            "astro.config.ts",
+            "astro.config.cjs",
+        ),
+        script_pattern=r"(?:^|\s)astro(?:\s|$)",
+    ),
+    source_extensions=(".astro",),
+)
+
+
+__all__ = ["ASTRO_SPEC"]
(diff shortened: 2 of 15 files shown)


elfensky:
### CI status note

This PR ships with 5 small test fixes that aren't part of the feature work. Here's why.

## The problem

When PR #614 was first opened, CI ran red on 7 tests. 5 of those were **failing on `main` itself** — they had been red on every PR since 2026-04-06, blocking every green build.

Two distinct root causes for the 5:

**3 bash tests** — `test_bash_unused_imports.py::test_bash_*_source_directive_*` and `test_bash_source_extra_arguments_are_not_imports`.

These call `detect_unused_imports`, which parses via tree-sitter. The `tests-core` CI job doesn't install `tree-sitter-language-pack`, so the function returns `[]`. The tests then assert that specific imports were detected, and fail because nothing was.

**2 review tests** — `test_review_commands.py::TestCmdReviewPrepare::test_do_run_batches_dry_run_generates_packet_and_prompts` (unit + integration).

These use `list(runs_dir.glob("*/prompts/batch-*.md"))` without sorting. Glob order is filesystem-dependent: on macOS the unsorted glob happens to return `batch-1.md` first; on Linux it returns `batch-2.md` first. The assertion `"Previously flagged issues" in prompt_text` then fails on Linux, because that string only appears in `batch-1.md` (the batch that has `historical_issue_focus`).

## The solution

**3 bash tests:** add a module-level skip-when-tree-sitter-missing mark, matching the existing pattern in `desloppify/tests/lang/common/test_treesitter.py`:

```python
pytestmark = pytest.mark.skipif(
    not is_available(),
    reason="tree-sitter-language-pack not installed",
)
```

**2 review tests:** wrap the glob in `sorted(...)`:

```python
prompt_files = sorted(runs_dir.glob("*/prompts/batch-*.md"))
```

These are bundled into the second commit on this PR (`chore(ci): unblock CI by gating bash tests on tree-sitter; sort review prompt glob`) so the PR can actually go green. Both fixes are independent of the Astro/Svelte/Vue work and could be cherry-picked back to `main` separately.

## Evidence these were pre-existing

1. Latest `CI` run on `main` ([25816808379](https://github.com/peteromallet/desloppify/actions/runs/25816808379)) failed on **exactly the same 5 tests**.
2. `gh run list --branch main --workflow CI` showed the last 3 push-triggered CI runs on `main` all failed, going back to 2026-04-06.
3. The first CI run on this branch ([26572311102](https://github.com/peteromallet/desloppify/actions/runs/26572311102), original PR commit, before any of my edits) and the latest pre-fix run had **identical failure sets** for these 5 — this PR's own edits did not introduce them.

## The PR's own 2 failures (now fixed)

The 2 dep-graph tests this PR adds (`test_dep_graph_treats_astro_frontmatter_as_importer`, `test_dep_graph_treats_mjs_config_as_importer`) were also failing in CI for the same tree-sitter reason. They're now gated with `@requires_treesitter`, so they run in `tests-full` (where tree-sitter is installed) and skip in `tests-core`.

elfensky:
### Round 2 review follow-up (commit `28ffdcb0`)

After the FrameworkSpec registry migration, ran a second multi-LLM review (Codex + Gemini + Claude + Perplexity). Five verified findings addressed in one commit. Full suite: **5820 passed, 6 skipped**.

**V1 — `vite` removed from Svelte/Vue `script_pattern`**
`specs/svelte.py:23`, `specs/vue.py:25` — Vite is a generic bundler, so a plain Vite+React project was matching both Svelte and Vue framework detection. Framework-specific tokens (`svelte-kit`, `nuxt`, `vue-cli-service`) already disambiguate.

**V2 — Framework scan now honors `opts.exclude`**
`treesitter/imports/graph.py`, `generic_support/registration.py` — the host `file_finder` was built via `make_file_finder(path_extensions, opts.exclude)`, but `_add_framework_importers` called `find_source_files` directly with no options. Introduced a `framework_file_finder: Callable` parameter through `make_ts_dep_builder` → `ts_build_dep_graph` → `_add_framework_importers`. `registration.py` now supplies a finder built with the same `opts.exclude` as the host. Framework files in user-excluded paths (e.g. `examples/`, `e2e/`) no longer contribute spurious importer edges.

**V3 — Commented-out imports no longer counted as live**
`treesitter/imports/graph.py:_strip_inline_comments` — strips `//`, `/* */`, and `<!-- -->` regions from each `grep_files` row before the import-specifier regex runs. Without this, `// import './foo'` in an Astro frontmatter or `<!-- import 'bar' -->` in a Vue template created false-positive importer edges and masked genuinely orphan modules — the orphan detector's primary failure mode. This was the strongest finding from **round 1** as well (2 providers agreed independently).

**V4 — Regression tests for missed shapes**
`tests/lang/common/test_treesitter_imports_direct.py` — three new tests:
- `test_framework_pass_ignores_commented_imports` — exercises all three comment styles
- `test_framework_pass_supports_vue_and_type_imports` — `.vue` `<script>` block + `import type {X} from 'y'` (neither was tested before, even though `.vue` is in the default tuple)
- `test_framework_file_finder_honors_caller_supplied_exclusions` — proves V2's exclude propagation

Plus `test_repeated_calls_return_cached_tuple` in `test_framework_source_extensions.py` for V5.

**V5 — `framework_source_extensions()` memoized**
`frameworks/registry.py` — per-ecosystem cache with `register_framework_spec` as the invalidation hook. `restore_registry` teardown bypasses `register_framework_spec` via direct dict mutation, so the fixture was updated to call `_invalidate_extensions_cache()` explicitly.

---

#### Findings discarded after verification (not acted on)

- **"`fw_abs[filepath]` will KeyError if `grep_files` normalises paths differently"** — invalid. `grep_files` (`grep.py:49`) returns the filepath strings verbatim from its input list; the lookup cannot KeyError. This was also flagged in round 1 and verified false then.
- **"Framework importer edges skipped unless target in `file_set`"** — this is the design. Graph nodes = host-language files only; framework files contribute one-way importer edges to host-language nodes. The PR's docstring at `_add_framework_importers` makes this explicit.
- **"`frameworks=True` not applied universally"** — verified deliberate. Only `javascript/__init__.py` uses the generic helper with `frameworks=True`; TypeScript has its own `deps/__init__.py` that calls `_framework_extensions()` directly. No latent gap.
- **"`ecosystem='node'` hardcoded in `_resolve_generic_extractors`"** — latent only; no non-node language currently sets `frameworks=True`. Worth a follow-up for future-proofing but not blocking.

PR #684, opened 2026-07-31
fix: keep rejected plan resolves transactional

ApexWorm:
## Problem

`desloppify plan resolve` wrote a `done` execution-log entry before the generic resolver enforced queue order. An out-of-order request could therefore leave the plan claiming completion while the issue remained open in state.

## Fix

Remove the premature wrapper write so the generic resolver is the first persistence path after all resolve guards pass. Add a regression using real state and plan files to prove a rejected out-of-order resolution leaves both unchanged.

## Verification

- `python3 -m pytest -q desloppify/tests/commands/plan desloppify/tests/commands/test_queue_order_guard.py`
- `python3 -m ruff check desloppify/app/commands/plan/override/resolve_cmd.py desloppify/tests/commands/plan/test_plan_override_transactions.py desloppify/tests/commands/plan/test_plan_overrides_direct.py desloppify/tests/commands/plan/test_workflow_gates.py`

### DIFF
--- a/desloppify/app/commands/plan/override/resolve_cmd.py
+++ b/desloppify/app/commands/plan/override/resolve_cmd.py
@@ -3,7 +3,6 @@
 from __future__ import annotations
 
 import argparse
-import logging
 
 from desloppify.app.commands.helpers.attestation import (
     show_attestation_requirement,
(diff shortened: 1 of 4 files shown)


awdemos:
Verified against #686: this PR's change is **exactly contained** in #686 — the new regression test file is byte-identical in both (`94474475…`), and the `resolve_cmd.py` diff is #684 plus #686's additions on top. They merge cleanly together (this becomes a no-op).

The fix itself is correct and well-tested — removing the premature `done` execution-log write so an out-of-order rejection leaves plan+state unchanged, with a real-file regression. No action needed from you; just flagging that if #686 lands, this can close as superseded (or, if #686 stalls, this is the clean version to land on its own).

PR #702, opened 2026-08-11
fix: stop reporting file-based-router entry points as orphaned

MercurialUroboros:
Withdrawn.

The measurements behind this were taken from a private codebase, and publishing its
shape was not something I had the standing to do. Body redacted for that reason.

The underlying issue is still real and worth fixing by someone: the `orphaned`
detector recognises only Next.js App Router conventions, so on any other
file-based-router framework it reports the entire routing surface as dead code —
nothing imports a route that the router loads by path. A second, separate cause is
that when a framework writes its `compilerOptions.paths` into a generated,
gitignored tsconfig, alias resolution falls back to a default that matches nothing,
so aliased imports look unresolvable.

Happy to reopen against a public fixture if a maintainer wants it.

### DIFF
--- a/desloppify/engine/detectors/orphaned.py
+++ b/desloppify/engine/detectors/orphaned.py
@@ -7,78 +7,10 @@ from collections.abc import Callable
 from dataclasses import dataclass
 from pathlib import Path
 
-from desloppify.base.discovery.file_paths import rel
-from desloppify.base.discovery.file_paths import count_lines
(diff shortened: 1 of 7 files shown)


MercurialUroboros:
Closing this — the measurements came from a private codebase and I'd rather not have its shape published. Happy to reopen a version built on a public fixture if the maintainers want the fix.

PR #704, opened 2026-08-12
fix: accept Codex batch JSON with token footer

### LINKED ISSUE #698: fix: accept Codex batch JSON followed by token footer
Running `desloppify review --run-batches --runner codex --parallel --scan-after-import` on macOS repeatedly fails after Codex returns a valid JSON review object followed by a `tokens used` footer. The run log records `Runner exited 0 but output file is missing or invalid; treating as execution failure.` The final JSON object is complete and valid before the footer, so the batch-result parser should either invoke Codex in JSON-only mode or extract and validate the final JSON object from stdout before rejecting the run.

szupzj18:
## Summary

- accept a valid Codex batch JSON object followed by the `tokens used` footer
- add regression coverage for the footer format

## Root cause

The runner validated the entire output with `json.loads()`, although the downstream batch parser already accepts a leading JSON object with trailing text. A valid Codex result was therefore reported as missing or invalid.

## Validation

- RED: footer regression test failed before the fix
- GREEN: 76 related tests passed
- Ruff critical checks passed

Fixes #698

### DIFF
--- a/desloppify/app/commands/review/runner_process_impl/io.py
+++ b/desloppify/app/commands/review/runner_process_impl/io.py
@@ -37,7 +37,8 @@ def _output_file_has_json_payload(output_file: Path) -> bool:
     if not output_file.exists():
         return False
     try:
-        payload = json.loads(output_file.read_text(encoding="utf-8", errors="replace"))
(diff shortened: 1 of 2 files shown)


szupzj18:

PR #706, opened 2026-08-12
fix: fail empty plan skip

### LINKED ISSUE #674: plan skip exits 0 on a zero-match pattern, silently dropping the skip
## Problem

`desloppify plan skip --permanent <pattern> ...` prints `No matching issues found` and **exits 0**. A caller scripting a triage stage has no way to distinguish "skipped 12 issues" from "skipped nothing", so a mistyped or superseded issue id silently no-ops.

## How it bit

During an organize stage assigning 232 review issues to clusters-or-skips, one issue carried a `superseded` marker that put it outside the open pool. `plan skip` reported no match and returned 0. Because the exit status was clean, the orchestration treated the skip as applied. It was only caught by re-auditing all 85 skips directly against `plan.json` afterwards.

The validation requirement for that stage is that *every* issue ends up clustered or skipped, so a silent no-op here produces an invalid triage that still reports success.

## Expected

Non-zero exit when a pattern matches nothing (or at minimum a `--strict` flag), so `set -e` orchestration fails loudly. A count of what was actually skipped in the success output would also let callers assert.

## Repro

```
desloppify plan skip --permanent "definitely-not-a-real-issue-id" \
  --note "x" --attest "I have reviewed this triage skip against the code and I am not gaming the score by suppressing a real defect."
echo "exit=$?"   # prints exit=0
```

szupzj18:
Fixes #674

Root cause: `plan skip` printed a warning and returned normally when no issue matched, so automation observed exit 0.

Change: raise `CommandError` with exit code 1 and cover the no-match path.

Validation: 58 related tests passed.

### DIFF
--- a/desloppify/app/commands/plan/override/skip.py
+++ b/desloppify/app/commands/plan/override/skip.py
@@ -222,8 +222,7 @@ def cmd_plan_skip(args: argparse.Namespace) -> None:
     plan = load_plan(plan_file)
     issue_ids = resolve_ids_from_patterns(state, patterns, plan=plan)
     if not issue_ids:
-        print(colorize("  No matching issues found.", "yellow"))
-        return
+        raise CommandError("No matching issues found.", exit_code=1)
 
     _warn_or_block_bulk_skip(issue_ids, confirm=bool(getattr(args, "confirm", False)))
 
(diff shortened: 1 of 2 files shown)


szupzj18:

PR #712, opened 2026-08-21
fix: scale and correct Rust coverage mapping

cjorge-graphops:
## Problems

1. Rust test-coverage import resolution rebuilt `RustProductionFileIndex` for every `use` specification. On a 648-production-file / 52-test-file workspace, coverage ran for more than 30 minutes at roughly 53,000 `newfstatat` calls per second.
2. Literal `include!("path.rs")` ownership was absent from the Rust dependency graph and direct-test mapping. Textually included implementation files were reported as zero-importer, untested modules even though they compile inside a directly tested owner.

## Fix

- Cache the immutable production-file index by resolved project root and sorted production scope.
- Resolve literal Rust source `include!` edges in the dependency graph.
- Expand direct-test ownership recursively through literal includes, matching Rust compilation semantics.
- Add focused cache, graph, and recursive coverage regressions.

## Verification

- Focused Rust/coverage tests: 25 passed.
- Full suite before the include change: 5,659 passed, 151 skipped; 5 unrelated baseline failures in Bash unused-import and review-prompt tests.
- Reproduction workspace: coverage completed in about 30 seconds instead of exceeding 30 minutes. Orphaned Rust files fell from 197 to 1, coverage findings fell from 306 to 158, and the full scan completed successfully.

### DIFF
--- a/desloppify/engine/detectors/test_coverage/detector.py
+++ b/desloppify/engine/detectors/test_coverage/detector.py
@@ -18,7 +18,7 @@ from .discovery import (
     _no_tests_issues,
     _normalize_graph_paths,
 )
-from .heuristics import _has_inline_tests
+from .heuristics import _expand_direct_test_targets, _has_inline_tests
 from .issues import (
     _generate_issues,
 )
@@ -71,6 +71,11 @@ def detect_test_coverage(
         )
     if test_files:
         directly_tested |= naming_based_mapping(test_files, production_files, lang_name)
+    directly_tested |= _expand_direct_test_targets(
+        directly_tested,
+        production_files,
+        lang_name,
+    )
 
     transitively_tested = transitive_coverage(directly_tested, graph, production_files)
     test_quality = analyze_test_quality(test_files, lang_name)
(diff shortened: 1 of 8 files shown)


awdemos:
Reviewed against `main` alongside #722 — the two PRs are competing implementations of the same perf bug (the `RustProductionFileIndex` being rebuilt for every `use` specification). Verdict on the split:

**The `include!()` modeling is unique to this PR and correct** — dep-graph edges plus recursive direct-test ownership for textually included files matches Rust's semantics, fixes zero-importer orphan reporting, and has no equivalent in #722 (#722's `promote_owner_covered_files` is a directory-convention heuristic covering a different case). That commit is worth landing.

**The cache commit is superseded by #722**, which is more complete: it also caches `describe_rust_file` and the workspace package index, and eliminates the per-file `get_test_files_for_prod` loop, which this PR leaves in place (cheapened, not removed). The two PRs also textually conflict in `languages/rust/tests/test_test_coverage.py`, and both ownership mechanisms would double-fire if stacked.

Recommend: drop the cache commit, rebase the `include!` commit onto whichever perf fix lands (#722's Rust-only commits, once split out of that bundle), and undraft. Happy to re-review quickly after that — the `include!` work is the kind of precise fix this detector needs.

Issue #422, opened 2026-03-14
Windows Codex batch runner breaks with codex 0.27.0


LostDaoWalker:
## Repro

Windows 11 + Python 3.13 + `desloppify 0.9.9` + `codex 0.27.0`

Run:

```powershell
desloppify review --run-batches --runner codex --parallel --scan-after-import
```

## Actual

The Codex batch runner fails before subjective review can complete.

Observed failures:

1. The runner invokes `codex exec --ephemeral ...`, but current Codex CLI rejects `--ephemeral`:
   - `error: unexpected argument '--ephemeral' found`

2. On Windows, the runner passes the full prompt as a command-line argument, which overflows command length:
   - `The command line is too long.`

3. Log/payload recovery uses locale-default text decoding on Windows, which crashes on UTF-8 output:
   - `UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d ...`

4. Live subprocess stdout/stderr decoding also uses locale-default decoding, which can crash the stream reader and then break the Codex child process with broken pipe.

## Expected

The Codex runner should:
- use the current Codex CLI contract
- feed prompts through stdin (or a file) instead of command-line args
- read logs/payloads with explicit UTF-8 decoding on Windows
- decode live subprocess streams with explicit UTF-8 as well

## Local fix that worked

I patched my local install to:
- remove `--ephemeral`
- call `codex exec ... --output-last-message <file> -`
- pipe the prompt via stdin
- set `encoding="utf-8", errors="replace"` for `subprocess.run` / `Popen`
- read batch logs and payload files with explicit UTF-8

After that, single-batch review runs successfully instead of failing at startup.

peteromallet:
Fixed in 0.9.10 — this overlaps with #442, both addressed by PR #453. The fix removes --ephemeral, pipes prompts via stdin instead of command-line args, and uses explicit UTF-8 decoding for logs and subprocess streams on Windows. Thanks for the report!

PR #718, opened 2026-08-23
fix(runner): decode live Codex subprocess output as UTF-8

### LINKED ISSUE #717: Regression: v1.0 live Codex subprocess decoding is locale-dependent on Windows
## Summary

`desloppify 1.0` regressed part of the Windows UTF-8 fix from #422 / #442. The persisted log readers use explicit UTF-8, but live Codex subprocesses in `runner_process_impl/attempts.py` still use `text=True` without `encoding` or `errors` in both `subprocess.Popen` and `subprocess.run`.

On a cp1252 Windows locale, Codex output can therefore raise `UnicodeDecodeError` in the stream reader before the already-safe log recovery runs.

## Current code

`_start_runner_process` and `_run_via_subprocess` pass `text=True`, but omit:

```python
encoding="utf-8",
errors="replace",
```

## Expected

Both live subprocess paths should decode Codex output explicitly as UTF-8 with replacement, matching the safe handling already used by `extract_payload_from_log` and `runner_failures.py`.

## Verified workaround

Adding `encoding="utf-8", errors="replace"` to both subprocess calls lets all Codex review batches run successfully on the same machine.

## Environment

- Windows 11, cp1252 system locale
- Python 3.14.7
- desloppify 1.0
- Codex CLI 0.149.0

szupzj18:
## Root cause

Fixes #717.

The live Codex subprocess paths — `_start_runner_process` and `_run_via_subprocess` in `runner_process_impl/attempts.py` — used `text=True` without `encoding`/`errors`. Output is therefore decoded with the locale encoding (`cp1252` on Western Windows), and UTF-8 Codex output containing non-ASCII characters raises `UnicodeDecodeError` in the stream reader before the already-safe log recovery runs.

The persisted-log readers (`io.py`, `runner_failures.py`) already pin `encoding="utf-8", errors="replace"` — the live subprocess paths were the gap (a regression of the #422/#442 UTF-8 work).

## Change

Pin `encoding="utf-8"`, `errors="replace"` on both subprocess calls, matching the persisted-log readers.

## Validation

- Two new unit tests assert both call sites request UTF-8 decoding (RED without the fix — `encoding` was `None`).
- Existing review fakes updated to accept the new kwargs.
- Full review suite green: `1081 passed`.

### DIFF
--- a/desloppify/app/commands/review/runner_process_impl/attempts.py
+++ b/desloppify/app/commands/review/runner_process_impl/attempts.py
@@ -125,6 +125,8 @@ def _start_runner_process(
             stderr=subprocess.PIPE,
             stdin=subprocess.PIPE if stdin_pipe else None,
             text=True,
+            encoding="utf-8",
+            errors="replace",
             bufsize=1,
         )
     except OSError as exc:
@@ -294,6 +296,8 @@ def _run_via_subprocess(
             run_kwargs = {
                 "capture_output": True,
                 "text": True,
+                "encoding": "utf-8",
+                "errors": "replace",
                 "timeout": deps.timeout_seconds,
             }
             if stdin_text is not None:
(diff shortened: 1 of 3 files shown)


szupzj18:

Issue #715, opened 2026-08-23
Tree-sitter dependency graph drops relative paths and CommonJS require edges


VGPlays4132:
### Desloppify version

1.0 (also present on current `main` as of 2026-08-23)

### Reproduction

Build a JavaScript dependency graph for project-relative files, for example:

```js
// app.js
const values = require('./lib/values');
```

`find_source_files()` returns project-relative graph keys (`app.js`, `lib/values.js`). `resolve_js_import()` also returns a project-relative path, but `ts_build_dep_graph()` converts the resolved path to absolute and compares it with the relative `file_set`. The edge is therefore discarded. In a CommonJS project, `JS_SPEC.import_query` additionally recognizes only ESM `import_statement`, not literal `require()` calls.

### Actual behavior

Every JavaScript file can be reported with zero importers. This produces large false-positive `orphaned` and test-coverage backlogs; in one project, 193 files were initially marked orphaned.

### Expected behavior

Resolved paths and discovered file keys should be compared in the same canonical path space, while graph output retains the original key spelling. Literal relative `require('./module')` calls should be included in the JavaScript/TypeScript import query.

### Verified local fix

- Build an absolute, normalized lookup from each discovered graph key, canonicalize resolver results through it, and add edges using the matching original key.
- Add a tree-sitter `call_expression` pattern constrained to identifier `require` with a literal string argument.

After these changes, the same project dropped from 193 to 38 zero-importer files and extracted modules showed their actual 1-12 importers.

PR #720, opened 2026-08-24
fix: recognize CommonJS require calls in JS/TS import graph

szupzj18:
## Problem

`JS_SPEC.import_query` (and `TYPESCRIPT_SPEC`) only matches ESM `import_statement`, so a CommonJS `require('./module')` call produces no import edge. Files imported only via `require` are reported with zero importers and show up as false `orphaned` findings. This is the second half of #715; #696 fixes the relative/absolute path-space mismatch in the same graph builder.

## Fix

Add a second pattern to the JS/TS import query that matches `require()` calls, using an `#eq?` predicate so other calls with string arguments are not treated as imports:

```
(call_expression
    function: (identifier) @require_fn
    arguments: (arguments (string (string_fragment) @path))
    (#eq? @require_fn "require")) @import
```

`resolve_js_import` already skips bare package names, so `require('lodash')` does not create an edge.

## Test

`test_js_dep_graph_includes_commonjs_require` builds a graph for a file with a relative `require`, a package `require`, and a non-require call, and asserts only the relative edge is kept. Fails without the query change, passes with it.

### DIFF
--- a/desloppify/languages/_framework/treesitter/_specs.py
+++ b/desloppify/languages/_framework/treesitter/_specs.py
@@ -691,6 +691,10 @@ JS_SPEC = TreeSitterLangSpec(
     import_query="""
         (import_statement
             source: (string (string_fragment) @path)) @import
+        (call_expression
+            function: (identifier) @require_fn
+            arguments: (arguments (string (string_fragment) @path))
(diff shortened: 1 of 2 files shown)


szupzj18:

PR #721, opened 2026-08-24
fix: harden TypeScript scanning and plan resolution

GDownes:
## Problem

Four scanner/workflow edge cases surfaced in a TypeScript monorepo:

- unused detection selected the repository-root tsconfig instead of the nearest config owning the scanned project;
- async function body extraction treated object-shaped parameter and return types as executable bodies;
- resolving one queued finding could auto-complete unrelated cluster steps, including steps whose triage refs were summary hashes rather than canonical issue IDs;
- temporarily skipped subjective reassessment prompts still selected the assessment lifecycle phase, then disappeared from display and masked executable triaged findings behind an empty queue.

## Fix

- create the temporary unused-check config beside the nearest owning tsconfig;
- scan balanced TypeScript signature delimiters before extracting the executable body;
- evaluate step completion against the whole living plan and fail closed while an unmatched-ref cluster still has members;
- remove skipped subjective items before phase resolution so queued implementation work remains executable.

## Validation

- 36 focused queue-snapshot tests passed for the latest fix; the earlier scanner and plan-resolution focused suites also pass.
- Ruff checks and format checks passed for the original six changed files.
- Full suite: 5,662 passed, 152 skipped; three pre-existing Bash source-directive tests fail in this environment.
- Verified all fixes against the WorkOrderGuard `services/api` scan and living plan; the previously empty queue now exposes the eight triaged review findings and planned coverage work.

### DIFF
--- a/desloppify/engine/_plan/step_completion.py
+++ b/desloppify/engine/_plan/step_completion.py
@@ -4,14 +4,20 @@ from __future__ import annotations
 
 
 def auto_complete_steps(plan: dict) -> list[str]:
-    """Mark steps done when all their issue_refs are no longer in the queue.
+    """Mark steps done when all their issue_refs leave the living plan.
 
     Returns list of human-readable messages for completed steps.
     """
     messages: list[str] = []
-    queue_set = set(plan.get("queue_order", []))
+    actionable_ids = set(plan.get("queue_order", []))
+    actionable_ids.update((plan.get("skipped") or {}).keys())
+    actionable_ids.update(plan.get("promoted_ids") or [])
+    for cluster in (plan.get("clusters") or {}).values():
+        if isinstance(cluster, dict):
+            actionable_ids.update(cluster.get("issue_ids") or [])
 
     for name, cluster in plan.get("clusters", {}).items():
+        cluster_issue_ids = set(cluster.get("issue_ids") or [])
         for i, step in enumerate(cluster.get("action_steps") or []):
             if not isinstance(step, dict) or step.get("done"):
                 continue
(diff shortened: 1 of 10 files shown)


awdemos:
Mixed bundle — two genuinely good fixes, one verified regression, one unsubstantiated claim, some scope creep. Detailed review against `main`:

**Keep:**
- `step_completion.py` fail-closed — real bug, reproduced on `main` (a step with summary-hash refs auto-completes immediately despite open cluster members); the new test correctly fails on main. One caveat for maintainer sign-off: the `all_gone and cluster_issue_ids: continue` brush also keeps canonically-referenced steps open until the whole cluster drains — deliberate but broad.
- `detector_core._extract_function_body` rewrite — broader than #713's async fix (covers arrow functions too, all 4 call sites); index alignment sound. 4 of its 5 new tests fail on main.

**Fix or drop:**
- `unused.py _find_base_tsconfig`: **verified regression** — on a Vite-style root with a `{"files": []}` solution `tsconfig.json` plus a real `tsconfig.app.json`, this picks `./tsconfig.json` (tsc compiles nothing → silent empty unused report) where main and #713 pick `./tsconfig.app.json`. It also keeps main's silent-empty behavior for tsconfig-less projects that #713 fixes. Recommend taking #713's variant for `unused.py` (per-directory `tsconfig.app.json` preference + source fallback) and keeping yours for everything else.
- `snapshot.py` skipped-subjective filter: the claimed bug **does not reproduce on main** — the new test passes pre- and post-patch (verified twice, with assessed and unassessed dimension fixtures). A test that passes before and after is not a regression test; make it fail on main or drop the hunk.
- `expectTypeOf` coverage pattern is fine but unrelated to the stated topic — 4–5 topics in one PR touching risky plan/queue core is scope creep.

Note this textually conflicts with #713 in `unused.py` — the two PRs fix the same two bugs with different designs (#721 wins on async extraction, #713 wins on tsconfig). They don't stack as-is; consolidating per above is the cleanest path.

PR #722, opened 2026-08-25
fix: bound Rust coverage dependency resolution

cjorge-graphops:
## Problem

Rust scans did not apply runtime/config exclusions while discovering Cargo packages, so ignored nested checkouts leaked into dependency resolution. Test coverage also rebuilt the full production-file index for every import and barrel candidate, turning a ~1,000-file project scan into a 20+ minute metadata-stat loop.

## Fix

- discover Cargo manifests through the exclusion-aware source walker
- key the package-index cache by runtime exclusions
- build direct test mappings in one pass instead of production×test resolution
- reuse cached Rust file, workspace, and production indexes across import and barrel resolution
- match module candidates against lexical/resolved in-memory indexes without filesystem probing
- add regressions for excluded workspaces, one-pass mapping, and shared production-index reuse

## Verification

- focused Rust/coverage suite: 103 passed
- real Autopilot V2 scan: all 13 phases complete; coverage falls from 20+ minutes to seconds
- full upstream suite: 5,795 passed, 151 skipped; 5 unrelated failures reproduce at upstream HEAD (three Bash unused-source cases and two review prompt cases)
- Ruff and diff checks pass on changed paths

### DIFF
--- a/desloppify/app/commands/plan/triage/confirmations/enrich.py
+++ b/desloppify/app/commands/plan/triage/confirmations/enrich.py
@@ -14,8 +14,7 @@ from .shared import (
     finalize_stage_confirmation,
 )
 from ..services import TriageServices, default_triage_services
-from ..stages.helpers import scoped_manual_clusters_with_issues
-from ..review_coverage import active_triage_issue_ids
+from ..stages.helpers import active_triage_issue_scope, scoped_manual_clusters_with_issues
 from ..validation.enrich_quality import (
     EnrichQualityIssue as _ConfirmationCheckIssue,
     EnrichQualityReport as _ConfirmationCheckReport,
@@ -179,7 +178,7 @@ def confirm_enrich(
     checks = _collect_enrich_level_confirmation_checks(
         plan,
         include_stale_issue_ref_warning=True,
-        triage_issue_ids=active_triage_issue_ids(plan, state) or None,
+        triage_issue_ids=active_triage_issue_scope(plan, state),
     )
 
     print(colorize("  Stage: ENRICH — Make steps executor-ready (detail, refs)", "bold"))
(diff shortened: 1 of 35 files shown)


awdemos:
The titled Rust-coverage work is real and well built — verified: exclusion-aware manifest discovery keyed by `get_exclusions()`, `describe_rust_file`/production-index/package-index caching, one-pass `build_production_test_index` replacing the per-production-file `get_test_files_for_prod` loop (quadratic pattern confirmed on main). 499 targeted tests pass on the PR vs 480 on main, zero regressions.

The problem is the bundle: **8 commits spanning at least 6 unrelated concerns**, with a body that mentions only the Rust part. Also included:

- empty-frozen-triage-scope semantics (core plan/triage, 5 files)
- `resolve_issues(status="fixed")` accepting `auto_resolved` (core state engine)
- trusted-import gate blocking only on *selected* dims (review command)
- glob support in `engine/policy/zones.py` (core policy)
- `boilerplate_min_lines=8` for Rust
- future-proofing skipped for `publish = false` crates

Each piece is individually reasonable and tested, but three of them **reduce findings** (owner-boundary promotion transitive→direct, zone reclassification of `*_tests/*`, boilerplate floor, publishable-only future-proofing) — the body's "coverage findings fell from 306 to 158" is partly reclassification, not just real coverage. Finding-reduction needs to be a disclosed, reviewable change, not ride-along hunks in a perf PR.

Recommend splitting: (1) Rust coverage perf PR, (2) triage/state semantics PR, (3) zones glob PR. Also note textual conflicts: with #712 in `languages/rust/tests/test_test_coverage.py` (competing implementations of the same perf fix — #712's `include!` handling has no equivalent here), and with #686 in 7 triage files.

PR #723, opened 2026-08-31
fix(cxx): cppcheck include-path harvesting and information-record classification

karabelaselias:
## Problem
On C++ projects whose sources include headers outside the scanned tree (vendored or sibling-package code), the cppcheck phase produced unactionable findings:
1. `_cppcheck_include_args` derived `-I` dirs only from in-scope headers, so external include roots recorded in the project's own `compile_commands.json` were never passed to cppcheck — every external header emitted `missingInclude` records.
2. cppcheck's `information` category (missing include paths, checker reports, normalization notes) is non-actionable by definition but was captured as findings.

## Fix
- Harvest `-I` paths from `compile_commands.json` entries matching the scanned files, falling back gracefully when the database is absent or malformed.
- Classify `information`-severity records as non-actionable in the phase filter.

## Verification
- New unit tests: information-record classifier, compile-db harvesting, malformed-database handling.
- Full upstream suite: 5818 passed, 3 skipped.
- Real-world check: C++17 rod-solver repo (114 files) — 47 information-level records and recurring missingInclude noise dissolved after the fix.

### DIFF
--- a/desloppify/languages/cxx/phases.py
+++ b/desloppify/languages/cxx/phases.py
@@ -3,6 +3,7 @@
 from __future__ import annotations
 
 import hashlib
+import json
 from pathlib import Path
 import shlex
 
@@ -18,7 +19,7 @@ from desloppify.languages._framework.generic_parts.parsers import PARSERS
 from desloppify.languages._framework.generic_parts.tool_factories import _record_tool_failure_coverage
 from desloppify.languages._framework.generic_parts.tool_runner import run_tool_result
 from desloppify.languages.cxx._helpers import build_cxx_dep_graph
-from desloppify.languages.cxx.extractors import find_cxx_files
+from desloppify.languages.cxx.extractors import CXX_HEADER_EXTENSIONS, find_cxx_files
 from desloppify.state_io import Issue
 
 CXX_COMPLEXITY_SIGNALS = [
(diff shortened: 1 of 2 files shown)


awdemos:
The core fix is sound and I verified the important part end-to-end: cppcheck `information:` records keep their severity prefix through `parse_gnu` (the parser only strips `file:line`), so `_is_information_record` is live production code — both information records dropped, real error/warning findings kept, and the header-suffix guard correctly spares a `.cpp` "is never used" entry. Malformed/absent `compile_commands.json` is handled gracefully (+3 tests pass).

Four things to address before merging:

1. **The body is inaccurate**: it claims `_cppcheck_include_args` already existed and "derived -I dirs only from in-scope headers" — on `main` that function does not exist and cppcheck gets *zero* `-I` args. The function is new in this PR. Please correct the description.
2. **Unadvertised flag changes**: `--check-level=exhaustive` requires cppcheck ≥2.16 — older distro cppcheck rejects it, so every batch errors and the detector yields zero findings (only reduced-coverage metadata) for those users. Also `--language=c++ --std=c++17` is forced on `.c` files, and `--inline-suppr` is added. None of these are mentioned in the body; each needs justification or removal.
3. **Unadvertised `_is_header_unused_diagnostic` filter** — also swallows `unusedFunction` "… is never used." findings located in headers. Over-suppression; at minimum disclose it.
4. **New ruff I001 import-sort error** (likely red in CI). Nit: relative `-I` paths in the compile db are resolved without the entry's `directory` field.

PR #734, opened 2026-09-07
Add Luau language plugin; fix import-graph and tool-output path handling

noahrepublic:
## Problem

Scanning a Roblox/Luau project reports a clean bill of health that nothing backs up. `.luau` is not registered against any language, so auto-detection falls through to `bash`, every mechanical detector runs over zero files, and the objective score comes back 100/100:

```
$ desloppify scan --path .
  Capabilities: linting (shellcheck), ...
  [7/12] Security...
         security: clean (0 files scanned)
  [9/12] Test coverage...
         test coverage: clean (0 production files)
  Scores: overall 25.0/100  objective 100.0/100  strict 25.0/100
```

The project has 857 `.luau` files. `--lang lua` does not help — it reports `No lua source files found`.

Adding `.luau` to the existing `lua` plugin's extensions is not sufficient: the Lua grammar produces ERROR nodes for type annotations, generics and `export type`, so AST-backed detection would degrade silently on any ordinarily-typed Luau codebase. `tree-sitter-language-pack` already ships a dedicated `luau` grammar.

## Fix

**New `luau` plugin** (`desloppify/languages/luau/`) on the `luau` grammar, with:

- A `resolve_luau_import` for require-by-string: `./x`, `../x`, `@self/x`, `.luaurc` aliases, and host aliases such as Roblox's `@game/...`. Host aliases address a runtime instance tree via a build manifest (Rojo) rather than the filesystem, so their tails are matched against a module index — longest tail first, accepting only an unambiguous match or, when a tree is duplicated, the copy nearest the requiring file. Ambiguity yields no edge rather than a guessed one.
- Both readings of a relative require inside an `init` file, since runtimes disagree on whether the base is the file or the directory it initialises. Only an existing file is accepted, so the ambiguity costs a stat rather than a wrong edge.
- Zone rules for vendored package trees and `.spec`/`.story` files, and entry patterns for `.client.luau`/`.server.luau` (started by the runtime, never required) and `Controllers/`/`Services/` (bulk-required at boot — the Roblox counterpart of the `/autoload/` entry in the GDScript plugin).
- `selene` as the linter. `stylua` is deliberately not wired: `--check` emits a diff, not diagnostics.

**Three framework bugs found while validating it**, each affecting more than Luau:

1. **`ts_build_dep_graph` dropped every edge when the file list was relative.** It normalised each resolved import to an absolute path, then tested membership against a file list that language file finders often return relative to the process cwd, so the check never matched. Edges are now matched on a normalised key that accepts either form, recorded under the original file-list spelling, and a self-import no longer makes a file its own importer. On the test project this went from **0 to ~1200 edges**, which in turn un-broke orphan, single-use and coupling detection.

2. **External tool findings bypassed the scan's exclusion rules.** Linters read their own config, not desloppify's, so running one at the project root reports on vendored trees the user excluded. On the test project **3154 of 4548 findings** came from a Wally `Packages/_Index/` directory that `desloppify exclude` had already ruled out. `make_tool_phase` now filters entries through the same exclusion patterns.

3. **`run_tool_result` decoded tool output with the locale codec.** A non-ASCII byte raised `UnicodeDecodeError` inside the subprocess reader thread and killed it mid-scan:

   ```
   Exception in thread Thread-5 (_readerthread):
     File "encodings\cp1252.py", line 23, in decode
   UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 4870080
   ```

   Output is now decoded as UTF-8 with `errors="replace"`.

Also adds `.luau` to the extension list the unused-import basename helper strips.

## Result on the test project

| | before | after |
|---|---|---|
| files scanned | 0 | 808 |
| import-graph edges | 0 | ~1200 |
| findings | 0 (fake 100/100 objective) | 1522 real |

## Tests

18 new tests (`test_luau_plugin.py`, plus dep-graph and tool-exclusion regressions). Full suite run before and after on the same machine: **35 failures both times** — all pre-existing and Windows-specific, none introduced — with passes going 5776 → 5794.

Three test doubles in `test_next_lint.py` had signatures pinned to the exact kwargs of `subprocess.run` and were widened with `**_kwargs`.


noahrepublic:
Pushed a fourth fix found while validating the plugin against a real project.

**Unused-import names came from the module path, not the binding.** That holds where the statement dictates the binding (`import foo`), but not where an import is an ordinary expression assigned to an ordinary variable — the norm in Lua/Luau:

```lua
local TeamClass = require("Tools/Team")      -- binds TeamClass, not Team
local MODULES = { require("Entities/awp") }  -- binds nothing at all
```

Both were reported unused. On the test project **all 31 unused-import findings were false**, 22 from one table of inline requires.

JavaScript already had a binding-aware path for this reason; this generalises it as an optional `import_binding` spec hook and wires it up for Luau. The hook returns the declaring statement as well as the name — the captured import node covers only the call, so excluding it alone leaves `local X = ` in the searched text, where the name matches itself and nothing ever looks unused.

Underscore-prefixed bindings are skipped, matching the Lua convention selene's `unused_variable` rule already honours.

Suite still at 35 pre-existing failures; passes now 5776 → 5799 (23 new tests).

awdemos:
Solid work — the plugin is genuinely well-built (`require` query constrained with `#eq?`, resolver handles relative/`@self`/`.luaurc`/host-alias tails with ambiguity-rejection, 18 tests) and the three bundled framework fixes are each real with regression tests. Mergeable as one PR after three small things:

1. **Rebase needed**: textual conflict with #699 and #703 in the shared `unused_imports.py` (2 hunks, trivial keep-both resolution, but mandatory).
2. **3 new ruff errors** vs clean main: 2× I001 unsorted imports (including the visibly misplaced `base.discovery` imports in `tool_factories.py`) and B905 `zip()` without `strict=` in `_shared_prefix_length`.
3. **`reset_luau_import_caches` is never wired into `reset_script_import_caches`** (which `scan/workflow.py` calls per scan). Unlike the PHP caches that are wired in, Luau's per-path module index can go stale across in-process rescans. One-line fix.

Heads-up on behavior worth a maintainer nod: `RES_SCRIPT_LITERAL_RE` treats *any* `res://….gd` string literal as reachability — deliberately lenient (false negatives possible), which matches the stated goal.

PR #738, opened 2026-09-10
fix: strip markdown fences and prose from opencode batch payloads

dopamine-pixels:
## Problem
desloppify review --run-batches --runner opencode reports valid batches as failed when the model wraps its JSON output in markdown fences or prefixes it with prose. The run log shows "Runner returned 0 but output file is missing" and the batch lands in the failed list even though a complete, valid payload exists in the stream.

## Fix
Normalise model output before parsing in _persist_opencode_payload_text:

- Strip surrounding ` fences when present.
- Fall back to the outermost {..} slice when prose surrounds the JSON.

## Verification
- Added tests: fenced-stream recovery through 
un_opencode_batch and direct _extract_json_payload_text cases (fenced, prose-prefixed, non-JSON).
- pytest desloppify/tests/commands/review/test_review_runner_helpers_direct.py — 12 passed.
- Observed in a real 20-batch review run: batches whose output was fenced JSON were the only failures; with this fix the retry round recovered them.


desloppify/tests/commands/review/test_review_runner_helpers_direct.py
dopamine-pixels, line 1: Tiny coverage nit: the fenced tests look like they’re actually exercising the {...} fallback since there’s prose before the fence. Could be worth adding one pure json ... case so the fence-stripping branch gets hit directly.

PR #752, opened 2026-09-18
fix: support Copilot personal skills and explain project install failures

AashrayAnand:
## Problem
Running update-skill copilot from an unwritable working directory (including a Windows-mounted tools directory under WSL) produces an uncaught PermissionError. The project-local destination is not shown beforehand, making it easy to mistake this for Windows-versus-Linux environment detection. Global setup supports several assistants, but currently omits Copilot CLI.

## Changes
- Add Copilot to global setup: desloppify setup --interface copilot installs the bundled skill at ~/.copilot/skills/desloppify/SKILL.md using Path.home(), independent of CWD and DESLOPPIFY_ROOT.
- Keep YAML frontmatter first for the personal skill and reuse existing global discovery/version handling.
- Preserve the existing project-local Copilot target and root resolution; no OS guessing, privilege escalation, or fallback to another destination.
- Show the resolved project target before downloading and translate filesystem failures into CommandError with the original error, project-root guidance, and a global-setup alternative where supported.
- Document project versus personal scope and distinguish Copilot CLI from the VS Code overlay.

## Validation
73 targeted tests passed across test_setup.py, test_update_skill_cmd_direct.py, test_transitive_modules_update_skill.py, and test_search_versioning.py. Coverage includes personal installation with a simulated Windows-mounted CWD and DESLOPPIFY_ROOT, repeat setup, global discovery, project instruction preservation, mkdir/read/write permission failures, and nonzero CLI exit without a traceback. Targeted repository lint selection passed. Also installed and re-ran setup in a real Linux user home.

Copilot personal skill location reference: https://docs.github.com/en/copilot/how-tos/copilot-cli/customize-copilot/add-skills


AashrayAnand:
@peteromallet Let me know what you think but I noticed the personal installation setup didn't exist for copilot.

This will work for copilot set up on windows and on wsl/linux

